PR #26, opened 2022-09-22
run: make it possible to stop runs that constantly fail

piodul:
Currently, the loop in `WorkerContext::run_worker` doesn't allow the worker to quit in case the operation it tries to perform constantly fails. This commit modifies the logic so that it will stop when asked, even if it got stuck on an error.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -75,6 +75,11 @@ impl WorkerContext {
             .store(INVALID_OP_ID_THRESHOLD, Ordering::Relaxed);
     }
 
+    // Was the worker asked to stop?
+    pub fn should_stop(&self) -> bool {
+        self.operation_counter.load(Ordering::Relaxed) >= INVALID_OP_ID_THRESHOLD
+    }
+
     // Issues the next operation id. If the context got a signal to stop
     // the stress operation, it will return `None`.
     fn issue_operation_id(&self) -> Option<u64> {
@@ -109,6 +114,7 @@ impl WorkerContext {
                     Ok(ControlFlow::Continue(_)) => continue 'ops,
                     Ok(ControlFlow::Break(_)) => break 'ops,
                     Err(err) if trial_idx >= self.max_retries_per_op => return Err(err),
+                    Err(err) if self.should_stop() => return Err(err),
                     Err(_) => continue 'trying,
                 }
             }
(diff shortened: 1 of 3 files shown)


piodul:
@cvybhu ping

src/run.rs
```diff
@@ -441,4 +447,35 @@ mod tests {
...
+            Err(anyhow::anyhow!("fail"))
+        }
+    }
+
+    #[tokio::test]
+    async fn test_ask_to_stop_on_constant_failures() {
+        let sem = Arc::new(Semaphore::new(0));
+        let sem_clone = Arc::clone(&sem);
+
+        let cfg = make_test_cfg(move || AlwaysFailsOp(Some(sem_clone.clone())));
```
cvybhu, line 469: I think that this test also needs max_restries_per_op to be set:
```rust
cfg.max_retries_per_op = 1_000_000_000_000_000_000;
```
Without it the worker doesn't get stuck in a retry loop and the test passes.

src/run.rs
```diff
@@ -441,4 +447,35 @@ mod tests {
...
+        let concurrency = cfg.concurrency as u32;
+
+        let (ctrl, fut) = run(cfg);
+
+        // Wait until all ops got stuck in retry loop
+        let _ = sem.acquire_many(concurrency).await.unwrap();
+
+        // Ask to stop and make sure that the workload finishes
+        ctrl.ask_to_stop();
+        fut.await.unwrap_err();
```
cvybhu, line 479: We could add some sort of timeout so that the test fails instead of hanging.
Maybe something like:
```rust
tokio::time::timeout(tokio::time::Duration::from_secs(10) fut).await.expect("Test timed out").unwrap_err();
```
Or using the [`ntest`](https://github.com/becheran/ntest/tree/master/ntest_timeout) crate?

cvybhu (changes requested):

src/run.rs
piodul: Well, the purpose of this PR was to make sure that the test doesn't keep retrying if stopping the benchmark was requested. The branch `Err(_) if self.should_stop()` added in the PR should make sure of that.

src/run.rs
piodul: Good point, I'll use the `ntest` crate.

src/run.rs
cvybhu: I commented out the `Err(_) if self.should_stop()` and the test still passed.
With `max_retries_per_op` set the new code is actually needed for the test to pass.

cvybhu (changes requested):
There are conflicts with master.

I still think the test requires setting the max retry argument.

piodul:
> There are conflicts with master.

Clippy started complaining, too... I'll fix clippy first, then rebase.

piodul:
v2: rebased, the test now also sets `max_retries_per_op = u64::MAX` in config

piodul:
@cvybhu review ping

cvybhu (approved):

PR #27, opened 2022-09-22
Add histogram-related options

piodul:
This PR add support for the following scylla-bench options related to latency histograms:

- `-latency-type` - determines which type of latency should be printed to the standard output (raw or coordinated-omission fixed),
- `-measure-latency` - controls whether latency should be measured at all,
- `-hdr-latency-sig` - controls the precision of the latency histograms (significant figures),
- `-hdr-latency-units` - determines the units used to record the latency,
- `-hdr-latency-file` - name of the file to which latency information should be saved, in the HistogramLogWriter format. Results can be viewed using e.g. the following page: https://hdrhistogram.github.io/HdrHistogramJSDemo/logparser.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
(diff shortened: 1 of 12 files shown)


cvybhu (changes requested):
Looks good.
The coordinated-omission fixed latency is a really weird concept, why would anyone want to use it 0_o

cvybhu (approved):

PR #30, opened 2022-09-23
s-b: driver-side timeouts

piodul:
Currently, driver-side timeouts are implemented manually using `tokio::time::timeout`. The newest version of the scylla-rust-driver, `0.5.0`, supports driver-side timeouts through its API. This PR updates the rust driver version and then removes the old, hacky way of enforcing timeouts, replacing it with the kosher one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.56"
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -17,6 +17,15 @@ dependencies = [
  "memchr",
```
cvybhu, line 17: Should `Cargo.lock` be tracked by git?
We gitgnore it in `scylla-rust-driver`, but maybe it's good too make sure that everyone that uses these benchmarks uses the same version of libraries.

cvybhu (approved):

Cargo.lock
piodul: As a rule of thumb, projects which have binary artifacts do track `Cargo.lock`, and rust libraries do not: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

> maybe it's good too make sure that everyone that uses these benchmarks uses the same version of libraries.

I agree, so let's keep it. We will naturally update it from time to time so that new releases of cql-stress should receive updates for the dependencies.

PR #33, opened 2022-09-30
treewide: appease clippy

piodul:
The `bool::then_some` function got stabilized, and clippy started complaining about the usage of `bool::then` with lambdas that immediately return a value. This commit switches uses of `then` to `then_some` where applicable.


piodul:
@cvybhu review ping

cvybhu (approved):

PR #34, opened 2022-10-06
run: get rid of a per-operation allocation 

piodul:
Currently, Rust does not support async in traits natively and usually
the "async-trait" crate is used to make async methods work. The main
limitation of async-trait is that it requires boxing the future returned
from the async function - so, an allocation.

The `Operation::execute` method is async, so it incurs an allocation for
each operation executed in the benchmark. While the driver itself
performs multiple allocations when handling a single request, for some
reason the allocation related to boxing the future was prominent in
flamegraphs of CPU-bound workloads (it took about 10% of the time).

This commit refactors the Operation so that it is responsible for
running its own `execute` loop in `Operation::run`. In order to make
migration to the new pattern easy, a macro `make_runnable!` is
introduced which generates an implementation of `run` based on the
existing `execute` method. Meanwhile, `execute` is demoted to a
non-trait async method, which Rust supports without any boxing.

This isn't a perfect abstraction because we can't force `run`
implementations to use `make_runnable!`, but the performance gains are
worth it. I couldn't find a way how to make the abstraction more tight;
I tried introducing a generic method which takes an async callback, but
it's very hard (if possible at all) to spell out the types and lifetimes
properly when the async callback takes non-static arguments and borrows
them in the returned future.

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1,9 +1,10 @@
-use std::ops::ControlFlow;
 use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::Result;
 use tokio::time::Instant;
+
+use crate::run::WorkerSession;
 
 /// Defines the configuration of a benchmark.
 pub struct Configuration {
@@ -76,10 +77,20 @@ pub trait OperationFactory: Send + Sync {
     fn create(&self) -> Box<dyn Operation>;
 }
 
-/// Represents an operation which is repeatedly performed during the stress.
+/// Represents an operation which runs its own operation loop.
+/// Implementing this interface instead of Operation leads to more efficient
+/// code because Rust, for now, forces us to Box futures returned
+/// from Operation::execute. This trait only incurs one allocation
+/// per running worker.
 #[async_trait]
 pub trait Operation: Send + Sync {
-    /// Executes the operation, given information in the OperationContext.
+    /// Classes that implement this trait should have the following, non-trait
+    /// method defined:
+    ///
+    /// async fn execute(&mut self, ctx: OperationContext) -> Result<ControlFlow<()>>;
+    ///
+    /// and they should use make_runnable!(TraitName) macro to generate
+    /// the implementation of the run() method.
     ///
     /// The operation should behave deterministically, i.e. the same action
     /// should be performed when given exactly the same OperationContext.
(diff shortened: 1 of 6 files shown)


piodul:
@cvybhu or @wprzytula, wanna take a look?

piodul:
@cvybhu @wprzytula ping

wprzytula:
The idea of using a macro to generate a proper implementation is right, yet using a macro_rules for this is counter-intuitive. Therefore, I've reworked it to be an attribute macro, which should be put, as normally, above a struct's definition. See #35 
@piodul

src/configuration.rs
```diff
@@ -89,5 +100,31 @@ pub trait Operation: Send + Sync {
...
 }
+
+/// Implements Operation for a type which implements an execute method.
+/// Although we could put execute() into the Operation trait, doing what we
+/// are doing here has better performance because asynchronous traits require
+/// putting returned futures in a Box due to current language limitations.
+/// Boxing the futures imply an allocation per operation and those allocations
+/// can be clearly visible on the flamegraphs.
+#[macro_export]
+macro_rules! make_runnable {
```
cvybhu, line 113: I think this would be prettier as a `#[derive(Runnable)]` macro.
That would require more work, so I guess this can stay if there's no time to do it.

src/run.rs
```diff
@@ -84,39 +84,62 @@ impl WorkerContext {
...
+    fn new(context: Arc<WorkerContext>) -> Self {
+        Self {
+            context,
+            op_id: 0,
+            trial_idx: 0,
         }
+    }
+
+    // Should be called before starting an operation.
+    pub async fn start(&mut self) -> Option<OperationContext> {
```
cvybhu, line 107: I got confuse by this methods name.
Maybe it could be renamed to `start_operation` and `end_operation`?
Otherwise it looks like we start a `WorkerSession` a million times.

cvybhu (changes requested):

src/configuration.rs
wprzytula: I've already done that in another PR on top of this: #35 !

piodul: I'm sorry, I changed my mind about the procedural macro. Procedural macros need to be put in a separate crate and when we will publish cql-stress to crates.io we will have to publish the new crate, too - which will make maintenance harder. I don't think converting to a derive macro is worth it.

src/run.rs
piodul: Good point, renamed to `start_operation` and `end_operation`.

PR #43, opened 2023-08-10
Cassandra stress rng

muzarski:
# Motivation
Cassandra-stress validates the data (`read` workload) inserted during `write` workload. It's easily achieved because of the deterministic nature of c-s value generation logic. We want to follow this approach, and generate exact same values as c-s does.

# Changes
- Imported `java_random` crate - Rust implementation of `java.util.Random`.
- Introduced `java_generate::distribution` module that defines multiple distributions. Note that there will be more distributions implemented once we let the user to specify the distributions via CLI.
- Introduced `java_generate::values` module responsible for generating random values. Currently, we are only able to generate data of `blob` type (`Vec<u8>`, in Rust driver). In the future, once `user` (which is very flexible, and lets the user define custom schemas) command is available to the users, we will support generating the values of other types.


src/bin/cql-stress-cassandra-stress/java_generate/faster_random.rs
```diff
@@ -0,0 +1,494 @@
...
+    #[test]
+    fn faster_random_test() {
+        let values_seed_one = generate_n_samples_with_seed(100, 1);
+        let values_seed_deadcafe = generate_n_samples_with_seed(100, 0xdeadcafe);
+        let values_seed_max_i64 = generate_n_samples_with_seed(100, i64::MAX);
+        let values_seed_min_i64 = generate_n_samples_with_seed(100, i64::MIN);
+
+        // Values generated from cassandra-stress.
+        // Used Java's implementation of FasterRandom.
+        assert_eq!(
```
avelanarius, line 70: `cargo fmt` gone mad :disappointed:

src/bin/cql-stress-cassandra-stress/java_generate/distribution/fixed.rs
```diff
@@ -0,0 +1,36 @@
+use super::{AtomicDistribution, Distribution};
+
+/// Distribution that always returns fixed value.
+pub struct FixedDistribution {
```
avelanarius, line 4: Maybe add a link to the corresponding Java class?

avelanarius: (here and in other distributions too)

src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs
```diff
@@ -88,3 +90,16 @@ pub trait ValueGenerator {
...
+#[cfg(test)]
+mod tests {
+    use super::{blob::Blob, Generator, GeneratorConfig};
+
+    #[test]
+    fn generator_config_salt_test() {
+        let blob_gen = Blob::default();
+        let config = GeneratorConfig::new(String::from("randomstrC0"), None, None);
+        let gen = Generator::new(blob_gen, config);
+        assert_eq!(gen.salt, 5919258029671157411);
```
avelanarius, line 104: Maybe this is too verbose (you don't have to agree with this suggestion), but maybe there should be a comment that those hardcoded values are taken from running the corresponding code in Java.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/uniform.rs
```diff
@@ -0,0 +1,50 @@
...
+        let sample = self.rng.next_double();
+        sample * (self.upper + 1.0) + (1.0 - sample) * self.lower
+    }
+
+    fn set_seed(&mut self, seed: i64) {
+        self.rng.set_seed(seed as u64)
+    }
+}
+
+impl NonAtomicDistribution for UniformDistribution {}
```
avelanarius, line 50: Could a test be added here?

src/bin/cql-stress-cassandra-stress/java_generate/values/hex_blob.rs
```diff
@@ -0,0 +1,113 @@
...
+            None,
+            // By default the keysize=10
+            Some(Box::new(FixedDistribution::new(10))),
+        );
+        let hex_blob = HexBlob();
+        let mut gen = Generator::new(hex_blob, config);
+
+        // -pop seq=1..5
+        // Samples from this distrubtion are the seeds to the partition key generator.
+        let mut seq = SeqDistribution::new(1, 5).unwrap();
```
avelanarius, line 76: This length might be a bit too small to check a scenario where `while i < size` in `generate()` loops twice.

src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs
```diff
@@ -0,0 +1,99 @@
...
+        size_distribution.set_seed(seed);
+        self.rng.set_seed(!seed);
+        let size = size_distribution.next_i64() as usize;
+
+        let mut result = vec![0; size];
+        let mut i = 0;
+        while i < size {
+            let v = self.rng.next_i64().to_le_bytes();
+            let n = min(size - i, v.len());
+            result[i..i + n].clone_from_slice(&v[0..n]);
```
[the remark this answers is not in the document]
Lorak-mmk: It should also be faster as it won't 0-initialize whole vector

Lorak-mmk (changes requested):

src/bin/cql-stress-cassandra-stress/java_generate/values/hex_blob.rs
```diff
@@ -0,0 +1,191 @@
+use super::ValueGenerator;
+use crate::java_generate::distribution::Distribution;
+
+/// Based on c-s HexBytes generator.
+/// Used by c-s to generate blob partition keys.
+/// TBH, not quite sure why c-s won't use Bytes generator for this.
+pub struct HexBlob();
```
wprzytula, line 7: Nit: needless tuple struct; it can be a unit struct (i.e. without parentheses).

Cargo.toml
```diff
@@ -14,19 +14,22 @@ env_logger = "0.9.0"
...
+scylla = { version = "0.8.1", features = ["ssl"] }
 sha2 = "0.10"
 strum = "0.25.0"
 strum_macros = "0.25.1"
 thread_local = "1.1.4"
-tokio = {version = "1.15.0", features = ["full"]}# TODO: Include only necessary features
-tracing = {version = "0.1.35", features = ["log"]}
+tokio = { version = "1.15.0", features = [
+    "full",
+] } # TODO: Include only necessary features
```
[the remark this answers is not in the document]
muzarski: This change is not introduced by this commit (something happened that this line was just broken into multiple lines - probably `cargo fmt`). We should ask @piodul  what should be done with this TODO.

src/bin/cql-stress-cassandra-stress/java_generate/faster_random.rs
```diff
@@ -0,0 +1,494 @@
...
+        random.set_seed(seed);
+        (0..n).map(|_| random.next_i64()).collect::<Vec<i64>>()
+    }
+
+    #[test]
+    fn faster_random_test() {
+        let values_seed_one = generate_n_samples_with_seed(100, 1);
+        let values_seed_deadcafe = generate_n_samples_with_seed(100, 0xdeadcafe);
+        let values_seed_max_i64 = generate_n_samples_with_seed(100, i64::MAX);
+        let values_seed_min_i64 = generate_n_samples_with_seed(100, i64::MIN);
```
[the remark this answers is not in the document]
muzarski: That's a good point. I don't have a strong opinion on that, maybe let's see what others think about it.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/mod.rs
```diff
@@ -0,0 +1,32 @@
+pub trait Distribution: Send + Sync {
+    fn next_i64(&mut self) -> i64;
+    fn next_f64(&mut self) -> f64;
+    fn set_seed(&mut self, seed: i64);
+}
```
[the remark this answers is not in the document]
muzarski: The distinction is introduced for efficiency (and correctness) purposes. It's needed for the distributions that use `java_random::Random` under the hood (which is not atomic). 

For these distributions, we will implement `Distribution` in a way, that each implementation will have its own copy of `java_random::Random` (as a member in a struct).

As of `AtomicDistribution`, me and @avelanarius decided to go with the `thread_local` approach (which is much faster than wrapping with mutex). In this case, there will be a `thread_local` copy of `java_random::Random` used to implement `next_{i/f}64_atomic` methods.

Note that distributions can be shared between that threads, as well as used by a single tokio task. We can't always use `AtomicDistributions` - because of the `thread_local`s. Assume one tokio task running on thread X changes the state of the underlying RNG (e.g. `set_seed(0)`), then it needs to `await` something - we switch to other task on the same thread X. Then the second task calls `set_seed(1)`. It needs to `await` something. We switch back to the first thread - it expects to `seed == 0`, which is obviously not the case. We strongly rely on the deterministic nature of c-s value generation logic.

When it comes to the distributions that will be shared across all tokio tasks, their state cannot be manipulated with `set_seed` method. What is also important, is the fact that there will be only one instance of such distribution during program execution.

 Please, correct me if the scenario I provided can never happen. From what I understood, `tokio` uses a thread pool - which implies that multiple tokio tasks can run on the same thread.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/uniform.rs
```diff
@@ -0,0 +1,518 @@
...
+}
+
+impl Distribution for UniformDistribution {
+    fn next_i64(&mut self) -> i64 {
+        bound(self.lower as i64, self.upper as i64, self.next_f64() as i64)
+    }
+
+    fn next_f64(&mut self) -> f64 {
+        let sample = self.rng.next_double();
+        sample * (self.upper + 1.0) + (1.0 - sample) * self.lower
```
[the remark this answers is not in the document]
muzarski: It's based on apache's `UniformRealDistribution`: https://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/src-html/org/apache/commons/math3/distribution/UniformRealDistribution.html#line.240

I'll add comment about it.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/mod.rs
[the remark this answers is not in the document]
wprzytula: What I understood both before and after you wrote the above is that `AtomicDistribution` is suitable for multithreaded use (by avoiding data races, which is by the way obligatory in _safe Rust_ to implement `Send + Sync`), but offers limited functionality (no `set_seed()`) in order to prevent race conditions, which could lead to logical errors (by an interlaced execution that you described).

The question I had in mind was: can't we make `Distribution` delegate to `AtomicDistribution`'s methods for generating random numbers? It's invalid though, because `AtomicDistribution` requires `Send + Sync` and `Distribution` does not.

Many thanks for clarification. I think it may be a good idea to put the above consideration in a comment, placed next to `Distribution` and `AtomicDistribution` trait definition.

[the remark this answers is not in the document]
muzarski: I added a comprehensive explanation of why do we need to introduce such distinction. It's located in the comments, next  to `Distribution` and `AtomicDistribution`. It also explains how are both types of distributions used in the row generation logic.

muzarski:
v2:
- fixed review comments
- rebased on top of https://github.com/scylladb/cql-stress/commit/6bf2fc14c72929cfb6131df28af7d68cd7000b74

muzarski:
v3:
- fixed additional review comments

# Distribution
Before the changes, we would make a distinction between `Distribution` (methods accepting `&mut self`) and `AtomicDistribution` (methods accepting `&self`). However, after noting that [thread_local](https://docs.rs/thread_local/latest/thread_local/) crate exists, I realized that the distinction is no longer necessary.

Previously, we would use std `thread_local!` macro which creates a thread_local object with static storage duration (one object, created with new thread). The `thread_local` crate, however, lets us to define multiple objects per thread (lazily allocated). This way, we can prevent the data races between multiple tokio tasks running on the same thread.

I refactored the code so the `Distribution`'s methods accept `&self`. The distributions that make use of `java_random::Random` can now use `ThreadLocalRandom` object, which is a simple thread_local wrapper for `java_random::Random`.

# Bonus
Resolved the TODO regarding including features of tokio. Now, we include only necessary features (used by cql-stress), instead of including all of the features. This will probably be moved to the separate PR, but as @wprzytula mentioned, it's a great place to address this issue.

wprzytula (approved):

src/bin/cql-stress-cassandra-stress/java_generate/distribution/mod.rs
```diff
@@ -9,6 +18,33 @@ pub trait Distribution: Send + Sync {
...
+
+    fn get(&self) -> RefMut<'_, Random> {
+        self.rng
+            .get_or(|| {
+                RefCell::new(Random::with_seed(
+                    SystemTime::now()
+                        .duration_since(UNIX_EPOCH)
+                        .map(|duration| duration.as_millis() as u64)
+                        .unwrap_or_default(),
+                ))
```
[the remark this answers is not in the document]
muzarski: c-s generates deterministic values only when partition_key generator seeds are sampled from DETERMINISTIC distribution (i.e. `SeqDistribution`). This approach is usually taken when populating the database with write routine.

For example:
```
./cassandra-stress write n=100 -pop dist=SEQ(1..100)
```
will generate the exact same set of 100 rows each time it's called.

Now, the data can be validated with `read` routine, for example:
```
./cassandra-stress read n=100 -pop dist=SEQ(1..100)
./cassandra-stress read n=100 -pop dist=UNIFORM(1..100)
./cassandra-stress read n=100 -pop dist=GAUSSIAN(30..70)
```
All of these workloads will successfully validate the data (inserted to DB with the write workload mentioned above).
To be more specific: any read workload that samples pk seeds from the subset of (1..100) range will successfully validate the data.

However, note that executing the following comments may not (and in most of the cases will not) be successful:
```
./cassandra-stress write n=100 -pop dist=UNIFORM(1..100)
./cassandra-stress read n=100 -pop dist=UNIFORM(1..100)
```
It's because the partition_key seed distribution is not deterministic. But as I mentioned, it is almost always the case that we populate the DB using some deterministic distribution.

When it comes to `Generator`s (which use `Distribution`s under the hood) - they are explicitly  seeded during row generation. See the row generation logic introduced in the following PR: https://github.com/scylladb/cql-stress/pull/45/commits/866a29c1ae496cfba9b8689a120ebef4caef0d21#diff-65a79d7c79bfc82d1f21b385ad5a7d03c5b12a1fcbc03113e41ea2c14ccbe544R27. `RowGeneratror` samples the `pk_seed` from the distribution shared by all of the tasks/threads (the distribution specified by the user with `-pop dist=...` option). The row values are fully dependent on the `pk_seed` sampled from this distribution.

Answering your question - yes, such seeding will stay in tune with c-s seeds.

muzarski:
v4:
- Extracted `ThreadLocalRandom` introduction to separate commit.

src/bin/cql-stress-cassandra-stress/java_generate/values/hex_blob.rs
```diff
@@ -0,0 +1,191 @@
+use super::ValueGenerator;
+use crate::java_generate::distribution::Distribution;
+
+/// Based on c-s HexBytes generator.
+/// Used by c-s to generate blob partition keys.
+/// TBH, not quite sure why c-s won't use Bytes generator for this.
```
piodul, line 6: As I can see, `HexBlob` just generates a sequence of hexadecimal digits instead of random bytes from range 0-255.

src/bin/cql-stress-cassandra-stress/java_generate/values/hex_blob.rs
```diff
@@ -0,0 +1,191 @@
...
+        while i < size {
+            let mut value = identity_distribution.next_i64();
+            let mut j = 0;
+            while j < 16 && i + j < size {
+                let v = (value & 15) as i32;
+                result[i + j] = ((if v < 10 { '0' as i32 } else { 'A' as i32 }) + v) as u8;
+                // Unsigned right shift.
+                value = ((value as u64) >> 4) as i64;
+                j += 1;
+            }
```
piodul, line 35: A hexadecimal digit is equal to 4 bits. `value` has 64 bits, and 64 = 4 * 16.

This code just takes a i64 random number (`value`) and generates at most 16 hexadecimal digits from it: takes 4 lowest bits (`let v = (value & 0xF) as i32`), writes a hexadecimal digit and finally discards the lowest 4 bits, shifting to the right (`value = ((value as u64) >> 4) as i64`).

src/bin/cql-stress-cassandra-stress/java_generate/values/hex_blob.rs
```diff
@@ -0,0 +1,191 @@
...
+    fn generate(
+        &mut self,
+        identity_distribution: &mut dyn Distribution,
+        size_distribution: &mut dyn Distribution,
+    ) -> Self::Value {
+        let seed = identity_distribution.next_i64();
+        size_distribution.set_seed(seed);
+        let size = size_distribution.next_i64() as usize;
+
+        let mut result = vec![0; size];
```
piodul, line 21: Why not `let mut result = Vec::with_capacity(size);` and then `result.push()` the bytes? This will skip the zero initialization of the vector.

piodul (approved):

PR #45, opened 2023-08-18
c-s runtime: write and read scenarios

muzarski:
# Changes
- Cosmetic changes in the settings module that simplify the access to the parameters (from the main).
- Introduced `operation` module.
- Implemented row-generation logic, based on c-s logic.
- Implemented write and read operations.
- Implemented main logic responsible for running the actual runtime scenario.

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs
+++ b/src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs
@@ -1,5 +1,7 @@
 use std::cmp::min;
 
+use scylla::frame::response::result::CqlValue;
+
 use super::ValueGenerator;
 use crate::java_generate::distribution::Distribution;
 use crate::java_generate::faster_random::FasterRandom;
@@ -12,13 +14,11 @@ pub struct Blob {
 }
 
 impl ValueGenerator for Blob {
-    type Value = Vec<u8>;
-
     fn generate(
         &mut self,
         identity_distribution: &mut dyn Distribution,
         size_distribution: &mut dyn Distribution,
-    ) -> Self::Value {
+    ) -> CqlValue {
         let seed = identity_distribution.next_i64();
         size_distribution.set_seed(seed);
         self.rng.set_seed(!seed);
@@ -33,7 +33,7 @@ impl ValueGenerator for Blob {
             i += n;
         }
 
-        result
+        CqlValue::Blob(result)
     }
 }
 
@@ -43,12 +43,17 @@ mod tests {
         distribution::fixed::FixedDistribution,
         values::{Generator, GeneratorConfig},
     };
+    use scylla::frame::response::result::CqlValue;
 
     use super::Blob;
 
     /// Utility function that maps u8 vector values to i8 values.
-    fn to_vec_i8(v: Vec<u8>) -> Vec<i8> {
-        v.into_iter().map(|x| x as i8).collect::<Vec<i8>>()
+    fn to_vec_i8(v: CqlValue) -> Vec<i8> {
+        v.as_blob()
+            .unwrap()
+            .iter()
+            .map(|x| *x as i8)
+            .collect::<Vec<i8>>()
     }
 
     #[test]
--- a/src/bin/cql-stress-cassandra-stress/operation/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mod.rs
@@ -1,0 +1,21 @@
+mod row_generator;
+
+use anyhow::Result;
+use std::num::Wrapping;
+
+pub use row_generator::RowGeneratorFactory;
+use scylla::{frame::response::result::CqlValue, QueryResult};
+
+/// See https://github.com/scylladb/scylla-tools-java/blob/master/tools/stress/src/org/apache/cassandra/stress/generate/PartitionIterator.java#L725.
+fn recompute_seed(seed: i64, partition_key: &CqlValue) -> i64 {
+    match partition_key {
+        CqlValue::Blob(key) => {
+            let mut wrapped = Wrapping(seed);
+            for byte in key {
+                wrapped = (wrapped * Wrapping(31)) + Wrapping(*byte as i64);
+            }
+            wrapped.0
+        }
+        _ => todo!("Implement recompute_seed for other CqlValues"),
+    }
+}
--- a/src/bin/cql-stress-cassandra-stress/operation/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mod.rs
@@ -1,9 +1,11 @@
+mod read;
 mod row_generator;
 mod write;
 
 use anyhow::Result;
 use std::num::Wrapping;
 
+pub use read::ReadOperationFactory;
 pub use row_generator::RowGeneratorFactory;
 use scylla::{frame::response::result::CqlValue, QueryResult};
 pub use write::WriteOperationFactory;
@@ -20,4 +22,53 @@ fn recompute_seed(seed: i64, partition_key: &CqlValue) -> i64 {
         }
         _ => todo!("Implement recompute_seed for other CqlValues"),
     }
+}
+
+fn validate_row(generated_row: &[CqlValue], query_result: QueryResult) -> Result<()> {
+    let rows = match query_result.rows {
+        Some(rows) => rows,
+        None => anyhow::bail!("Query result doesn't contain any rows.",),
+    };
+
+    let first_row = match rows.split_first() {
+        Some((first_row, remaining_rows)) => {
+            // Note that row-generation logic behaves in a way that given partition_key,
+            // there is exactly one row with this partition_key.
+            anyhow::ensure!(
+                remaining_rows.is_empty(),
+                "Multiple rows matched the key. Rows: {:?}",
+                rows
+            );
+            first_row
+        }
+        None => anyhow::bail!("Query result doesn't contain any rows.",),
+    };
+
+    anyhow::ensure!(
+        first_row.columns.len() == generated_row.len(),
+        "Expected row's ({:?}) length: {}. Result row's ({:?}) length: {}",
+        generated_row,
+        generated_row.len(),
+        first_row.columns,
+        first_row.columns.len(),
+    );
+
+    let result =
+        first_row
+            .columns
+            .iter()
+            .zip(generated_row.iter())
+            .all(|(maybe_result, expected)| match maybe_result {
+                Some(result) => result == expected,
+                // TODO: For now, we don't permit NULLs.
+                None => false,
+            });
+
+    anyhow::ensure!(
+        result,
+        "The data doesn't match. Result: {:?}. Expected: {:?}.",
+        first_row.columns,
+        generated_row,
+    );
+    Ok(())
 }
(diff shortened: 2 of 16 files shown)


src/bin/cql-stress-cassandra-stress/settings/command/help.rs
```diff
@@ -13,24 +13,24 @@ pub fn print_help() {
     Options::print_generic_help();
 }
 
-fn print_command_help(command_str: &str) -> Result<CommandParams> {
+fn print_command_help(command_str: &str) -> Result<Option<CommandParams>> {
     match Command::parse(command_str) {
         Ok(cmd) => {
             cmd.print_help();
-            Ok(CommandParams::Special)
+            Ok(None)
```
piodul, line 20: > This commit introduces very subtle changes in the settings module.
> It helps the users of the settings access the configuration in a more
> convenient way.

I'd appreciate if the first part of the commit message were less vague. A commit should concentrate on doing one thing, and "very subtle changes in the settings module" is very permissive with regard to what can happen there.

It looks like the idea is to replace `CommandParams` with `Option<CommandParams>` and get rid of `CommandParams::Special`, right? I think you should mention it. The justification that it's more convenient (idiomatic?) is already good.

Besides, a more specific description slightly helps me because I don't have to spend as much time trying to _guess_ the exact idea behind the commit, instead I can concentrate on _verifying_ whether your code aligns with your intentions.

src/bin/cql-stress-cassandra-stress/settings/command/help.rs
```diff
@@ -13,24 +13,24 @@ pub fn print_help() {
...
         )),
     }
 }
 
-fn print_option_help(option_str: &str) -> Result<CommandParams> {
+fn print_option_help(option_str: &str) -> Result<Option<CommandParams>> {
     Options::print_help(option_str)?;
-    Ok(CommandParams::Special)
+    Ok(None)
 }
```
piodul, line 31: I'm not sure if I understand why this function returned `Result<CommandParams>` in the first place. It is supposed to only print help, right? Returning `Result<>` I can understand, but `CommandParams`?

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
```diff
@@ -5,7 +5,7 @@ use crate::settings::{
 use anyhow::Result;
 
 pub struct RateOption {
-    threads_info: ThreadsInfo,
+    pub threads_info: ThreadsInfo,
```
piodul, line 8: Is this change supposed to be included in this commit?

src/bin/cql-stress-cassandra-stress/settings/mod.rs
```diff
@@ -22,7 +23,7 @@ use self::option::SchemaOption;
 
 pub struct CassandraStressSettings {
     pub command: Command,
-    pub params: CommandParams,
+    pub command_params: CommandParams,
```
piodul, line 26: This rename should probably be a separate commit. It looks like you really are mixing multiple things in the first commit, but you shouldn't.

src/bin/cql-stress-cassandra-stress/settings/command/read_write.rs
```diff
@@ -327,11 +327,16 @@ fn parse_with_handles(handles: ReadWriteParamHandles) -> ReadWriteParams {
...
+    cmd: &Command,
+    payload: &mut ParsePayload,
+) -> Result<Option<CommandParams>> {
     let args = payload.remove(cmd.show()).unwrap();
     let (parser, handles) = prepare_parser(cmd.show());
     parser.parse(args)?;
-    Ok(CommandParams::BasicParams(parse_with_handles(handles)))
+    Ok(Some(CommandParams {
+        basic_params: parse_with_handles(handles),
+    }))
```
piodul, line 339: This function always returns either `Err()` or `Ok(Some(...))`. Does it make sense for it to return `Result<Option<CommandParams>>` instead of `Result<CommandParams>`?

src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs
```diff
@@ -1,5 +1,7 @@
 use std::cmp::min;
 
+use scylla::frame::response::result::CqlValue;
```
piodul, line 3: > c-s values_generator: return CqlValue

Nit: typo in the commit message: `implementiation` -> `implementation`

src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
```diff
@@ -0,0 +1,147 @@
...
+/// - seed the `pk_generator` with sampled `pk_seed`
+/// - generate the partition key with `pk_generator`
+/// - compute the seed for the `column_generators` based on generated pk
+/// - generate the rest of the row (seeding the `column_generators` with computed seed)
+///
+/// I think it's a great place to address how read and write workloads cooperate.
+/// For reference, see: https://github.com/scylladb/cql-stress/pull/43#discussion_r1304274035.
+///
+/// At first, we need to notice that as long as `pk_seed_distribution` is a DETERMINISTIC distribution,
+/// the row-generation logic is also deterministic.
```
piodul, line 27: What is a deterministic distribution? Is this concept defined somewhere?

src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
```diff
@@ -0,0 +1,147 @@
...
+/// ./cassandra-stress read n=100 -pop dist=UNIFORM(1..100)
+/// ./cassandra-stress read n=100 -pop dist=GAUSSIAN(30..70)
+/// ```
+///
+/// To be more precise: any read workload that samples the partition_key seeds from the distribution
+/// sampling the values from the subset of range 1..100 (one used in the write routine) will successfully validate the data.
+///
+/// This also means, that if we used a NON-DETERMINISTIC distribution e.g. `UniformDistribution` (which is by default seeded with current time as millis)
+/// in the write workload, most of the times, the read workload will result in a validation error.
+/// For example:
```
piodul, line 50: Side note: it's technically possible to achieve determinism with non-seq distributions, and AFAIR we already do this in the s-b command - you can use the operation number as the seed to the random generator, and then use it to generate the partition key. I'm not sure why the original c-s doesn't do it...

Perhaps it's worth considering this as a future improvement.

piodul (changes requested):

src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
muzarski: Good point, and to be quite honest, I'd really love to follow the approach you mentioned. I'm not sure what c-s wanted to achieve by being non-deterministic in the first place.

I'll save this conversation for the future (in the comment), and hopefully we will discuss it with @avelanarius .

src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs
```diff
@@ -43,12 +43,17 @@ mod tests {
...
 
     /// Utility function that maps u8 vector values to i8 values.
-    fn to_vec_i8(v: Vec<u8>) -> Vec<i8> {
-        v.into_iter().map(|x| x as i8).collect::<Vec<i8>>()
+    fn to_vec_i8(v: CqlValue) -> Vec<i8> {
+        v.as_blob()
+            .unwrap()
+            .iter()
+            .map(|x| *x as i8)
+            .collect::<Vec<i8>>()
```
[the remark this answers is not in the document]
muzarski: It's a test-utility function. There were no changes to the test itself, that's why we don't see the usage in this commit.

src/bin/cql-stress-cassandra-stress/settings/command/mod.rs
```diff
@@ -77,31 +77,24 @@ impl Command {
...
-    // TODO: MixedParams,
-
-    // USER
-    // TODO: UserParams,
+pub struct CommandParams {
+    // Parameters shared across all of the commands
+    pub basic_params: ReadWriteParams,
+    // TODO:
+    // mixed_params: Option<MixedParams>
+    // user_params: Option<UserParams>
```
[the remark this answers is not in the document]
muzarski: If you look at the parameters of some commands, you can see that parameters of `read/write` are a subset of parameters of both `mixed` and `user` command. IMO, it makes sense to get rid of the enum, and represent it as struct, since both `mixed` and `user` will include `basic_params`, and some additional parameters specific to them.

src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs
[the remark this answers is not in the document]
piodul: Ah, OK. For some reason I thought that this is a new function, but indeed it is not.

src/bin/cql-stress-cassandra-stress/settings/command/mod.rs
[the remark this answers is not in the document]
piodul: Ok. Again, this should be mentioned in the commit message.

muzarski:
v2:
Fixed review comments, including (the most important):
- split the first commit into 3 commits
- extended commit message regarding `CommandParams` refactor

src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
muzarski: Added description in the comment.

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
muzarski: Extracted to yet another commit

muzarski:
v3:
Rebased on top of master

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -21,3 +23,41 @@ fn recompute_seed(seed: i64, partition_key: &CqlValue) -> i64 {
...
+            // there is exactly one row with this partition_key.
+            anyhow::ensure!(remaining_rows.is_empty(), "Multiple rows matched the key.",);
+            first_row
+        }
+        None => anyhow::bail!("Query result doesn't contain any rows.",),
+    };
+
+    anyhow::ensure!(
+        first_row.columns.len() == generated_row.len(),
+        "Generated row's length: {}, result row's length: {}",
```
piodul, line 45: How about printing those rows?

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -21,3 +23,41 @@ fn recompute_seed(seed: i64, partition_key: &CqlValue) -> i64 {
...
+            .columns
+            .iter()
+            .zip(generated_row.iter())
+            .all(|(maybe_result, expected)| match maybe_result {
+                Some(result) => result == expected,
+                // TODO: For now, we don't permit NULLs.
+                None => false,
+            });
+
+    anyhow::ensure!(result, "The data doesn't match.");
```
piodul, line 61: How about printing those rows so that the user can see how the mismatch looks like?

piodul (commented):
Some nitpicks related to error messages, but otherwise LGTM

piodul (approved):

PR #51, opened 2023-09-13
c-s: params wrapper

muzarski:
# Motivation

Before the changes, a lot of common logic (and fields) would be repeated in implementations of `Param` trait. I found it annoying when implementing yet another type of parameters (`DistributionParam` - https://github.com/scylladb/cql-stress/pull/48) to repeat the same boilerplate code.

Simple examples:
- `required`, `prefix`, `satisfied`, `desc` fields used by both `SimpleParam` and `MultiParam`
- getters and setters for the fields in `Param` trait - the exact same implementations for both types of parameters
- common parsing logic - e.g. check if user-provided argument starts with `prefix`.

This PR introduces a new generic type `ParamWrapperImpl<P: Param>` which wraps the `Param` implementations. The logic shared between all of the types of parameters is extracted to the wrapping type.

# Changes
- Introduced generic `ParamWrapperImpl<P: Param>` type which wraps the common logic of the parameters
- Introduced `ParamWrapper` trait implemented (once) by all of the generic `ParamWrapperImpl` types. This trait is needed so we can store both `ParamWrapperImpl<SimpleParam>` and `ParamWrapperImpl<MultiParam>` in common collections used by the parser.
- Extracted the shared logic from `Param` implementations to `ParamWrapperImpl`
- Changed `ParamCell` type used by the parser from `Rc<RefCell<dyn Param>>` to `Rc<RefCell<dyn ParamWrapper>>`. Adjusted usages of the type.

# Notes
Please review this PR before https://github.com/scylladb/cql-stress/pull/48, since the options implemented in the mentioned PR base on the changes introduced by this PR.

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/settings/param/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/param/mod.rs
@@ -12,13 +12,8 @@ pub use multi_param::MultiParamHandle;
 pub use parser::ParamsParser;
 pub use simple_param::SimpleParamHandle;
 
-/// An 'interface' of parameter.
-///
-/// Note that the parser uses trait objects.
-/// For now, it may seem to be unnecessary since we only support `SimpleParam`s.
-/// However, cassandra-stress supports more complex parameters (see help -schema)
-/// which cql-stress should support in the future as well.
-pub trait Param {
+/// A specific implementation of the parameter.
+pub trait ParamImpl {
     /// Checks whether `arg` matches parameter's prefix.
     fn try_match(&self, arg: &str) -> bool;
     /// Parses the `arg` value.
@@ -40,7 +35,7 @@ pub trait Param {
     fn print_desc(&self);
 }
 
-type ParamCell = Rc<RefCell<dyn Param>>;
+type ParamCell = Rc<RefCell<dyn ParamImpl>>;
 
 pub trait ParamHandle {
     fn cell(&self) -> ParamCell;
--- a/src/bin/cql-stress-cassandra-stress/settings/param/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/param/mod.rs
@@ -16,7 +16,6 @@ pub use simple_param::SimpleParamHandle;
 pub trait ParamImpl {
     /// Parses the `arg` value.
     fn parse(&mut self, arg: &str) -> Result<()>;
-    fn required(&self) -> bool;
     /// Ref: check `ParamsGroup`.
     /// Checking whether the group is satisfied happens right after all of the
     /// CLI arguments were successfully consumed. If the group is satisfied,
@@ -99,7 +98,7 @@ impl<P: ParamImpl> GenericParam for TypedParam<P> {
     }
 
     fn required(&self) -> bool {
-        self.param.required()
+        self.required
     }
 
     fn try_match(&self, arg: &str) -> bool {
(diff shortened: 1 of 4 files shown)


src/bin/cql-stress-cassandra-stress/settings/param/mod.rs
```diff
@@ -76,7 +76,36 @@ impl<P: Param> ParamWrapperImpl<P> {
...
+///   specific type of the parameter that is being wrapped. They hold the parameters in
+///   some collections (std::Vector) and use a common interface of the [Param].
+/// - On the other hand, the values returned to the user depend on the specific type of the parameter
+///   For example, [simple_param::SimpleParam] exposes `get()` method which returns a simple value,
+///   when [multi_param::MultiParam] exposes `get_arbitrary` method (returning arbitrary parameters as a map).
+///
+/// The implementation for this trait is written only once (below) for all of the generic types
+/// of [ParamWrapperImpl]. Thanks to this trait, [ParamWrapper]s can be held in a common collection
+/// as [Rc<RefCell<dyn ParamWrapper>>]. At the same time, corresponding handles (e.g. [simple_param::SimpleParamHandle])
+/// can access the specific type, exposing specific methods (e.g. [ParamWrapperImpl<simple_param::SimpleParam>]::get()).
```
piodul, line 87: I think this docstring comment should be more "straight to the point". It should explain in one sentence what the item is, then more specific explanations should follow. The current comment throws in lots of references to other items at the very beginning and seems to assume that the reader already knows about the rest of the code.

Take a look at the section about documenting components from the rustdoc book: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components . Maybe we don't necessarily need to go as far and provide code examples, but the general idea should be that you start with the most important information and explain more details as the docstring unfolds.

I'm also not sure that all the details about other structs/traits need to be included here. Such comments have a tendency to become obsolete or misleading if you restructure other parts of the code.

piodul (changes requested):
The naming of structs/traits in `cql-stress-cassandra-stress/settings/param/mod.rs` looks confusing to me. I suggest some changes.

First, the `ParamWrapper`/`ParamWrapperImpl` seem to be the objects that represent a parameter in some sense - know how to parse the parameter, keep its state etc. They are similar, but one is esentially a type-erased version of the other. I'm not sure what is the convention in such a case, but in the module for parsing Go flags I used `Generic-` and `Typed-` prefixes, so maybe we could use `GenericParam` and `TypedParam`?

As for the existing `Param`, the trait is essentially about customizing a param's behavior, so I propose to change it to `ParamFlavor`.

muzarski:
v2:
- renamed:
  - trait `ParamWrapper` -> `GenericParam`
  - struct `ParamWrapperImpl` -> `TypedParam`
  - trait `Param` -> `ParamImpl`
- thinned out docstring comments for the types. Especially removed the references to specific `ParamImpl` implementations in the comments
- Adjusted commit messages to the renamed types

muzarski:
v3:
rebased on top of master

muzarski:
@piodul should I squash some of the commits (e.g. porting fields from `ParamImpl` implementations to `TypedParam`) before mergning?

piodul:
> @piodul should I squash some of the commits (e.g. porting fields from `ParamImpl` implementations to `TypedParam`) before mergning?

No, I think the commits are fine as they are right now.

piodul (approved):

PR #53, opened 2023-09-20
cassandra-stress CI

muzarski:
# Motivation
Core logic of cassandra-stress frontend has already been implemented, but there are no tests verifying the correctness of our solution.

This PR introduces two tests written in python which test runtime logic of `cql-stress-cassandra-stress` tool.

# Changes
- fixed shebang in existing `tools/test_with_scylla.py` python script
- introduced `docker/scylla_test/compose.yml` docker-compose yaml defining a scylla node used during the tests
- extracted `ScyllaDockerNode`, as well as the logic corresponding to this type, to `tools/util/scylla_docker.py`
- `scylla_docker` function now makes use of `docker compose` to set up the container. Thanks to `docker-compose up --wait` flag, we can get rid of part of the code responsible for polling the scylla container until it's healthy.
- introduced `tools/util/cassandra_stress.py` which defines set of utility functions when working with cassandra-stress
- introduced two tests in `tools` directory. Included these tests in CI
- updated `README.md`


piodul (changes requested):

tools/util/cassandra_stress.py
```diff
@@ -0,0 +1,50 @@
...
+@contextmanager
+def cassandra_stress(cassandra_version="3.11.16"):
+    try:
+        # Fetch the cassandra
+        print(f"Fetching cassandra {cassandra_version}")
+        cassandra_dir = f"apache-cassandra-{cassandra_version}"
+        cassandra_tar = f"{cassandra_dir}-bin.tar.gz"
+        # See https://www.apache.org/dyn/closer.lua/cassandra/3.11.16/apache-cassandra-3.11.16-bin.tar.gz
+        cassandra_url = f"https://dlcdn.apache.org/cassandra/{cassandra_version}/{cassandra_tar}"
+        subprocess.run(args=["wget", "-N", cassandra_url], check=True, stderr=subprocess.DEVNULL)
```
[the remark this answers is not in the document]
muzarski: Looking at the output of some example CI run (e.g. https://github.com/muzarski/cql-stress/actions/runs/6248180046/job/16962268279) I decided to drop the output and not display it, since it looks kind of messy. What do you think about that?

And why stderr? Well, I was confused as well but apparently `wget` prints everything to stderr instead of stdout.

tools/util/cassandra_stress.py
```diff
@@ -0,0 +1,50 @@
...
+                             "-rate", "threads={}".format(threads), 
+                             "-schema", "keyspace={}".format(keyspace)],
+                             check=True)
+
+
+@contextmanager
+def cassandra_stress(cassandra_version="3.11.16"):
+    try:
+        # Fetch the cassandra
+        print(f"Fetching cassandra {cassandra_version}")
```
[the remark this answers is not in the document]
muzarski: Yes, I even considered to do so. I can add some CLI argument providing path the path to cassandra-stress tool.

I think we can follow the same logic we use during scylla docker container setup:
- if user specified scylla ip, we assume that the container is healthy, up and running (CI run case)
- otherwise, we set up the container ourselves (lazy repo user case)

We could do the same for fetching cassandra-stress. We can fetch cassandra in GH workflow, and then provide the path to cassandra-stress to test scripts. However, if someone wanted to run these scripts locally, we should fetch cassandra from python.

.github/workflows/rust.yml
```diff
@@ -37,3 +37,5 @@ jobs:
       run: cargo build --bin cql-stress-cassandra-stress
     - name: C-S frontend test 1
       run: ./tools/test_cs_write_and_validate.py --scylla-ip 127.0.0.1 --workload-size 100 --concurrency 1
+    - name: C-S frontend test 2
+      run: ./tools/test_cs_equal_db.py --scylla-ip 127.0.0.1 --workload-size 100 --concurrency 1
```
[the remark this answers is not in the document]
piodul: I would like for the developers to be able to quickly run the same tests locally as the ones that run on the CI. Now, you have to copy and paste commands from the workflow file into the terminal to run the same tests - I don't know a way to run a GitHub workflow locally.

tools/util/cassandra_stress.py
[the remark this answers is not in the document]
piodul: I think that you can use `-s` flag to silence this particular output. I'd rather not ignore stderr unconditionally in order not to skip some important messages (if any such messages can appear).

There is some other command lower here which redirects stdout - same concern there.

tools/test_cs_equal_db.py
```diff
@@ -0,0 +1,88 @@
...
+            print("\n=== cql_stress rows: ===\n")
+            for row in cqlstress_rows:
+                print(row)
+            raise AssertionError("Test failed. Tables are not equal.")
+        else:
+            print("\n=== Test successful ===\n")
+    finally:
+        # Drop keyspaces
+        session.execute(f'drop keyspace {ks_cassandra}')
+        session.execute(f'drop keyspace {ks_cqlstress}')
```
piodul, line 55: I think we shouldn't drop those keyspaces. Imagine a developer that runs some test locally, if there is data mismatch then it might be convenient for them to have this data in the database so that they can efficiently query it etc.

To solve the issue of keyspace/table reuse, we can generate keyspace names based on a random number + timestamp and also print it at the beginning of the test.

tools/test_cs_equal_db.py
```diff
@@ -0,0 +1,88 @@
...
+    print("\n=== Comparing tables... ===\n")
+    try:
+        if cs_rows != cqlstress_rows:
+            print("Compared tables are not equal!!!")
+            print("\n=== cassandra-stress rows: ===\n")
+            for row in cs_rows:
+                print(row)
+            print("\n=== cql_stress rows: ===\n")
+            for row in cqlstress_rows:
+                print(row)
```
piodul, line 48: I wonder whether we should always print this. In CI, we are only inserting a small number of rows, but developers might use a larger number of rows when running bigger tests locally.

tools/test_cs_write_and_validate.py
```diff
@@ -0,0 +1,83 @@
...
+    session = cluster.connect()
+
+    # All setup, start the test.
+    print("\n=== Starting the test... ===")
+
+    # Populate DB with cassandra-stress.
+    print("\n=== Populating database with cassandra-stress... ===\n")
+    cs.run(command="write",
+           workload_size=args.workload_size,
+           node_ip=node.ip, keyspace="ks_cassandra",
```
piodul, line 23: `ks_cassandra` and `ks_cqlstress` should be variables. Furthermore, I think we should generate unique names for those keyspaces and avoid dropping at the end of the test - see my other comments.

tools/test_cs_write_and_validate.py
```diff
@@ -0,0 +1,83 @@
...
+        # Use user-provided container.
+        node = ScyllaDockerNode(ip=args.scylla_ip, port="9042")
+        run(args, node, cs)
+    else:
+        # Create the container and use it for the test.
+        with scylla_docker(args.teardown) as node:
+            run(args, node, cs)
+
+
+
```
piodul, line 81: Nit: one blank line too much?

tools/util/cassandra_stress.py
```diff
@@ -0,0 +1,52 @@
...
+        self.stress_cmd = stress_cmd
+
+
+    def run(self, command, workload_size, node_ip, keyspace, threads):
+        args = [self.stress_cmd]
+        args += prepare_args(command, workload_size, node_ip, keyspace, threads)
+        subprocess.run(args=args, check=True)
+
+
+def cassandra_stress(cassandra_version="3.11.16"):
```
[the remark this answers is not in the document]
piodul: Besides, maybe it would make sense for the function to be the constructor of `CassandraStress` class?

tools/util/cassandra_stress.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Extract cassandra
+        print(f"Extracting cassandra {cassandra_version}")
+        subprocess.run(args=["tar", "-xzf", f"{download_dir}/{cassandra_tar}", "--directory", download_dir], check=True)
+        print(f"Extracted cassandra to {download_dir}/{cassandra_dir}")
+
+    return CassandraStress(stress_cmd=f"{download_dir}/{cassandra_dir}/tools/bin/cassandra-stress")
+
+
+def run_cql_stress(command, workload_size, node_ip, keyspace, threads):
```
piodul, line 49: It's a little inconsistent that we have `CassandraStress` object but no `CqlStress` object. What about introducing one?

tools/test_cs_equal_db.py
```diff
@@ -0,0 +1,88 @@
+#! /usr/bin/env python3
+
+
+from util.cassandra_stress import *
+from util.scylla_docker import *
```
piodul, line 5: Nit: switch from glob imports to explicit imports

piodul (changes requested):
I wonder whether we should use an existing test framework (e.g. `pytest`) for this. It's true that we don't have many tests right now (will we ever?) so I don't think we would gain much from it as developers, but maybe QA would be interested in running the tests and need output in a particular format?

cc: @avelanarius

tools/util/cassandra_stress.py
```diff
@@ -0,0 +1,52 @@
...
+    cassandra_tar = f"{cassandra_dir}-bin.tar.gz"
+
+    if os.path.exists(f"{download_dir}/{cassandra_dir}"):
+        # Cassandra already fetched.
+        print(f"Cassandra {cassandra_version} already installed. Skipping the download phase.")
+    else:
+        # Fetch cassandra.
+        print(f"Fetching cassandra {cassandra_version} to {download_dir}")
+        # See https://www.apache.org/dyn/closer.lua/cassandra/3.11.16/apache-cassandra-3.11.16-bin.tar.gz
+        cassandra_url = f"https://dlcdn.apache.org/cassandra/{cassandra_version}/{cassandra_tar}"
```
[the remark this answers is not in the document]
muzarski: > The comment points to a different domain than the `cassandra_url` variable - why?

It displays the recommended url to fetch cassandra from. But yeah, I'll change that so it's a direct url to cassandra.

> Besides, why use version `3.*` when `4.*` is available?

That's a good question. At first, I tried the latest version (being 5.* AFAIR), but the tests failed. The same happened when using version 4.*.

TLDR - Apache c-s changed its generation logic a bit (the commit - https://github.com/apache/cassandra/commit/f1f5f194620d3f9e11492f0051b6b71018033413). Notice, however, that `scylla-tools-java` still follows the generation logic used in 3.* versions of Apache Cassandra.

tools/test_cs_equal_db.py
muzarski: True that. I'll remove it, since we won't drop the keyspaces anyway.

tools/test_with_scylla.py
```diff
@@ -5,44 +5,20 @@
...
-def scylla_docker(teardown_behavior, max_healthcheck_attempts):
+def scylla_docker(teardown_behavior, compose_yaml="docker/scylla_test/compose.yml"):
     print("Setting up Scylla")
-    cp = subprocess.run(args=["docker", "run", "-d", "-p", "9042:9042", "scylladb/scylla"],
-                        capture_output=True, check=True)
-    container_id = cp.stdout.decode('utf-8').strip()
-    print(f"Container ID: {container_id}")
+    print("Starting the container and waiting for it to be healthy")
+
+    subprocess.run(args=["docker", "compose", "-f", compose_yaml, "up", "-d", "--wait"], check=True)
```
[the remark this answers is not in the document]
muzarski: When the container is not healthy:
I changed healthcheck command to `exit 1`, then started the container:

```
$ docker compose -f docker/scylla_test/compose.yml up -d --wait
[+] Running 2/2
 ✔ Network scylla_test_default  Created           0.1s 
 ✔ Container scylla_test        Started           0.0s 
container scylla_test is unhealthy
```
The exit code is 1 (subprocess will fail due to `check=True`).

Then running `docker ps`:
```
$ docker ps
CONTAINER ID   IMAGE             COMMAND                  CREATED          STATUS                      PORTS                                                                                             NAMES
67efb4d13951   scylladb/scylla   "/docker-entrypoint.…"   57 seconds ago   Up 57 seconds (unhealthy)   22/tcp, 7000-7001/tcp, 9160/tcp, 9180/tcp, 10000/tcp, 0.0.0.0:9042->9042/tcp, :::9042->9042/tcp   scylla_test
```

We see that container is still running, so need to proceed with the teardown.

muzarski:
# v3
## Changes
- applied `PEP-8` formatting to all of the python scripts
- replaced harcdoded keyspaces names with a generated ones
- changed the logic so we don't drop the keyspaces after the tests
- Introduced `CqlStressCassandraStress` class. It inherits from `CSCliRunner` along with `CassandraStress`
- added `cassandra-download` directory to `.gitignore`.
- log grouping prints in `./tools/cassandra-stress-ci.sh` are now printed only in the GH actions environment (when `CI` env variable is set).

## Notice
We still don't make use of the `pytest` framework. I need to get familiar with the framework, but don't want to delay the review process of the things that are already fixed.

muzarski:
v4:
rebased on top of master

muzarski:
v5:
using `pytest` to define test-cases

tools/cassandra_stress_ci.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Test cases
+
+
+def test_write_and_validate(default_runtime_args, scylla_docker_node,
+                            cassandra_stress, cql_stress):
+    start_ci_group("C-S test 1")
+    run_write_and_validate(runtime_args=default_runtime_args, node=scylla_docker_node,
+                           cs=cassandra_stress, cql_stress=cql_stress)
+    end_ci_group()
```
piodul, line 60: I think we should rather remove the `start_ci_group` and `end_ci_group` methods altogether. It will give weird results, e.g. pytest prints all logs from the failing tests at the end - so github would see them more than once.

Besides, side note for the future: if you want to make sure that something happens even if the test fails (i.e. `end_ci_group`), then it should happen in a teardown method (correspondingly, `start_ci_group` would be printed in setUp).

README.md
```diff
@@ -67,3 +67,40 @@ If you are using a non-standard IP address or port for your Scylla instance, you
...
+```
+
+To run the test cases used during CI (defined in `./tools/cassandra_stress_ci.py`), you can make use of [pytest](https://pytest.org).
+Before running the tests, make sure you have scylla up and running.
+If you use some non-standard scylla URI, you can specify it via `SCYLLA_URI` env variable.
+```bash
+docker compose -f docker/scylla_test/compose.yml up -d --wait
+export SCYLLA_URI="127.0.0.1:9042"
+pytest -s ./tools/cassandra_stress_ci.py
+```
```
piodul, line 106: The docs seem outdated. For example, there is no `test_cs_equal_db.py` script anymore.

piodul (changes requested):
We're getting there :) Left some comments, but after fixing them it should be GTM

piodul (approved):

avelanarius (approved):

Issue #54, opened 2023-09-21
Allow configuring the number of connections per shard


piodul:
It would be useful to test how Scylla behaves when there are lots of open CQL connections to each node. A rust driver session opens one connection per shard by default but it can be configured to open more, so this is just a matter of adding an appropriate commandline option.

The option should be added both to s-b and c-s frontends.

muzarski:
Can we close this issue?

piodul:
Yes. Next time, please add "Fixes: #54" (with an appropriate issue number, of course) to the description of the PR so that it is closed automatically.

PR #55, opened 2023-09-21
multiple shard connections

muzarski:
# Motivation

As mentioned in https://github.com/scylladb/cql-stress/issues/54, we should let the user configure the number of connections per shard.

This PR extends both s-b and c-s CLI by option which configures shard connection count.

# Changes
- extended s-b `ScyllaBenchArgs` by `shard_connection_count` field. Removed TODO comments and unused variables regarding `connections` option.
- extended c-s `NodeOption` by `shard_connection_count` field.


muzarski:
v2:
renamed s-b `shard-connection-count` CLI argument to `connection-count`, so the test passes

src/bin/cql-stress-scylla-bench/args.rs
```diff
@@ -100,6 +101,8 @@ where
         true,
         "use compression for client-coordinator communication",
     );
+    let shard_connection_count =
+        flag.u64_var("connection-count", 1, "number of connections per shard");
```
piodul, line 105: Let's change the name of the parameter to indicate that this is a per-shard connection count.

src/bin/cql-stress-cassandra-stress/settings/option/node.rs
```diff
@@ -103,6 +109,12 @@ fn prepare_parser() -> (ParamsParser, NodeParamHandles) {
         "Preferred datacenter for the default load balancing policy",
         false,
     );
+    let shard_connection_count = parser.simple_param(
+        "connections=",
```
piodul, line 113: Does the original c-s have an option like that? If so, I'd suspect that it specifies per-node connection count, not per-shard. I'd rather use a different name that suggests that it's a per-shard count.

piodul (changes requested):

src/bin/cql-stress-cassandra-stress/settings/option/node.rs
muzarski: No, there is no such option in the original c-s. I agree with the name change.

src/bin/cql-stress-cassandra-stress/settings/param/types.rs
```diff
@@ -41,6 +41,15 @@ impl Parsable for u64 {
     }
 }
 
+impl Parsable for NonZeroUsize {
+    type Parsed = NonZeroUsize;
+
+    fn parse(s: &str) -> Result<Self::Parsed> {
+        s.parse::<NonZeroUsize>()
+            .with_context(|| format!("Invalid usize value: {s}"))
+    }
```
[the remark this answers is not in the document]
muzarski:
```
$ ./target/release/cql-stress-cassandra-stress write n=10 -node connections=0
...

Failed to parse parameter connections=.

Caused by:
    0: Invalid usize value: 0
    1: number would be zero for non-zero type
Error: Failed to parse CLI arguments.
```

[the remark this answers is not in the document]
piodul: > `    0: Invalid usize value: 0`

`0` is a valid usize value. This error message looks strange.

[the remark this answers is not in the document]
muzarski: Should I change it to `Invalid non-zero usize value` or just not provide any additional context (and map the error type to `anyhow::Error`)?

src/bin/cql-stress-scylla-bench/args.rs
muzarski: What about the `connection-count` parameter (which was just a placeholder until now)? The commands in `args_test.in` include such parameter.

We can:
- either replace `connection-count` with `shard-connection-count`
- or include both of them (still leaving `connection-count` as a placeholder since original s-b supports such parameter)

muzarski: The other idea I have is to have a parameter which defines the number of connections, as well as a type (per-shard/per-node). So something like `--connection-count per-shard(10)` or `--connection-count per-node(10)`.

piodul: > What about the connection-count parameter (which was just a placeholder until now)? The commands in args_test.in include such parameter.

Let's introduce the new `shard-connection-count` and keep ignoring the old one.

> The commands in `args_test.in` include such parameter.

If we are going to replace s-b in scylla-cluster-tests, I'd rather not change the existing behavior of the tool, so I'd rather keep ignoring the existing option.

muzarski:
v3:
fixed review comments

piodul (approved):

PR #56, opened 2023-10-02
cassandra-stress - gaussian distribution

muzarski:
# Changes
Implemented `NormalDistribution`:
- added `Random` struct which wraps `java_random::Random`, in addition exposing `next_gaussian()`
- implemented gaussian distribution parsing logic (`NormalDistributionFactory`)
- added tests based on Java results


src/bin/cql-stress-cassandra-stress/java_generate/mod.rs
```diff
@@ -1,3 +1,56 @@
...
+    fn with_seed(seed: u64) -> Self {
+        Self {
+            delegate: java_random::Random::with_seed(seed),
+            next_next_gaussian: None,
+        }
+    }
+
+    fn set_seed(&mut self, seed: u64) {
+        self.delegate.set_seed(seed)
+    }
```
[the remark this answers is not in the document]
piodul: From the docs of java.util.Random.setSeed:

> ... The method setSeed is implemented by class Random by atomically updating the seed to
>
>  (seed ^ 0x5DEECE66DL) & ((1L << 48) - 1)
>
> and clearing the haveNextNextGaussian flag used by [nextGaussian()](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextGaussian--).

src/bin/cql-stress-cassandra-stress/java_generate/mod.rs
```diff
@@ -1,3 +1,56 @@
 pub mod distribution;
 pub mod faster_random;
 pub mod values;
+
+/// A structure wrapping [java_random::Random].
+/// In addition to methods exposed by [java_random::Random],
+/// it exposes [next_gaussian] method as well.
+struct Random {
```
piodul, line 8: > Notice: Java's implementation of this function makes use of `StrictMath`,
> which may give different results for square root and natural log functions for floats.
> However, the results shouldn't differ when we cast the them to integers.

I wouldn't be so sure. If the result of `next_gaussian()` is close to an integral value and both implementations return slightly different results - so that one returns a value below the integer and the other one above it - then you will get different results when casting to an integer.

How important it is to have exactly the same results as the Java's `Random`?

src/bin/cql-stress-cassandra-stress/java_generate/distribution/normal.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn verify_args(min: i64, max: i64, standard_deviation: f64) -> Result<()> {
+        anyhow::ensure!(
+            min < max,
+            "Upper bound ({}) for normal distribution is not higher than the lower bound ({}).",
+            max,
+            min
+        );
+        anyhow::ensure!(
+            standard_deviation > 0f64,
+            "Standard deviation must be posivite"
```
piodul, line 25: Nit: typo: `posivite` -> `positive`

src/bin/cql-stress-cassandra-stress/java_generate/distribution/normal.rs
```diff
@@ -56,3 +57,99 @@ impl Distribution for NormalDistribution {
...
+
+impl NormalDistributionFactory {
+    fn do_parse_from_description(desc: &Description<'_>) -> Result<Box<dyn DistributionFactory>> {
+        // See https://github.com/scylladb/scylla-tools-java/blob/master/tools/stress/src/org/apache/cassandra/stress/settings/OptionDistribution.java#L202.
+        let mut iter = desc.args.iter();
+
+        let (min, max) = match (iter.next(), iter.next()) {
+            (Some(min), Some(max)) => (min.parse::<i64>()?, max.parse::<i64>()?),
+            _ => anyhow::bail!("Expected at least 2 arguments"),
+        };
```
piodul, line 97: Nit: we already have `Description::check_argument_count`, so maybe we could have a method which checks that we have at least N arguments as well.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/normal.rs
```diff
@@ -56,3 +57,99 @@ impl Distribution for NormalDistribution {
...
+
+        let (mean, stdev) = match (iter.next(), iter.next(), iter.next()) {
+            (Some(mean), Some(stdev), None) => (mean.parse::<f64>()?, stdev.parse::<f64>()?),
+            (maybe_stdvrng, None, None) => {
```
piodul, line 101: In general, if one invocation of `iter.next()` returns `None`, the next one is _not guaranteed_ to return `None` again. See: https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next

That said, I think it _should_ be the case for `desc.args.iter()` as it iterates over a Vec (slice), but it's not immediately obvious which kind iterator is that and, for the sake of clarity, I'd [fuse](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fuse) it.

piodul (changes requested):

src/bin/cql-stress-cassandra-stress/java_generate/mod.rs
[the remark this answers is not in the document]
muzarski: Good catch, haven't thought about it.

src/bin/cql-stress-cassandra-stress/java_generate/mod.rs
muzarski: Yeah... I am aware of the fact that there is a non-zero probability that the samples from Rust and Java may differ (even though it's very small).

The only scenario (I can think of) where cql-stress might be inconsistent with c-s is:
```
cql-stress write n=10 -col 'size=GAUSSIAN(1..100)'
cassandra-stress read n=10 -col 'size=GAUSSIAN(1..100)'
```
Assume that within 10 [number of ops] * 5 [default number of blob columns in row] = 50 samples from the GAUSSIAN(1..100) distribution, there is some `i` such that `cql_stress_samples[i] != c_s_samples[i]`.
Then we get a cassandra-stress read error (data was not successfully validated).

Notice that `-col size=` (the distribution from which we sample the size of the blob cell) is the only distribution that can be set by the user, and in addition, samples from this distribution are deterministic (since we [seed](https://github.com/scylladb/cql-stress/blob/master/src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs#L23) the distribution before each sample).

The other distribution that we use in a deterministic way (set seed before each sample), is an `identity_distribution`. This distribution, however, has a default value of ['UNIFORM(1..100B)'](https://github.com/scylladb/cql-stress/blob/master/src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs#L56) and cannot be set by the user via CLI.

Now the question is, whether we want to import an additional crate with `StrictMath` (e.g. https://crates.io/crates/rust-strictmath), just for the sake of Gaussian distribution behaving in the same way as Java's implementation does.

NOTICE: this may also be the problem in the future once we decide to seed `pk_seed` distribution with operation's id. This distribution is set via `-pop dist=` parameter, and is used to sample the seeds for `partition_key` generator for each operation.

muzarski:
v2:
- fixed review comments
- added `Description::check_minimum_argument_count`
- added `Description::args_fused`

src/bin/cql-stress-cassandra-stress/java_generate/mod.rs
piodul: > Yeah... I am aware of the fact that there is a non-zero probability that the samples from Rust and Java may differ (even though it's very small).
> 
> The only scenario (I can think of) where cql-stress might be inconsistent with c-s is:
> 
> ```
> cql-stress write n=10 -col 'size=GAUSSIAN(1..100)'
> cassandra-stress read n=10 -col 'size=GAUSSIAN(1..100)'
> ```
> 
> Assume that within 10 [number of ops] * 5 [default number of blob columns in row] = 50 samples from the GAUSSIAN(1..100) distribution, there is some `i` such that `cql_stress_samples[i] != c_s_samples[i]`. Then we get a cassandra-stress read error (data was not successfully validated).

Isn't `[number of ops]` going to be much higher in practice, i.e. in SCT tests? If so, then the probability rises. If we are going to write with our implementation of c-s and then validate with the original c-s, then this is prone to flakiness.

> 
> Notice that `-col size=` (the distribution from which we sample the size of the blob cell) is the only distribution that can be set by the user, and in addition, samples from this distribution are deterministic (since we [seed](https://github.com/scylladb/cql-stress/blob/master/src/bin/cql-stress-cassandra-stress/java_generate/values/blob.rs#L23) the distribution before each sample).
> 
> The other distribution that we use in a deterministic way (set seed before each sample), is an `identity_distribution`. This distribution, however, has a default value of ['UNIFORM(1..100B)'](https://github.com/scylladb/cql-stress/blob/master/src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs#L56) and cannot be set by the user via CLI.
> 
> Now the question is, whether we want to import an additional crate with `StrictMath` (e.g. https://crates.io/crates/rust-strictmath), just for the sake of Gaussian distribution behaving in the same way as Java's implementation does.

If 100% compatibility with the output of the original c-s is required, then I guess that we should.

> 
> NOTICE: this may also be the problem in the future once we decide to seed `pk_seed` distribution with operation's id. This distribution is set via `-pop dist=` parameter, and is used to sample the seeds for `partition_key` generator for each operation.

muzarski: > Isn't `[number of ops]` going to be much higher in practice, i.e. in SCT tests? If so, then the probability rises. If we are going to write with our implementation of c-s and then validate with the original c-s, then this is prone to flakiness.

Yes, that was just an example. The biggest number of ops I've tried so far (locally) is 10M - everything went smoothly.

> If 100% compatibility with the output of the original c-s is required, then I guess that we should.

Maybe @avelanarius will have a stronger opinion on that. I am still not sure about that.

muzarski:
v3:
- rebased on top of master
- making use of strictmath library to generate `next_gaussian` values. Notice that `rust-strictmath` dependency points to the fork of the original repository, since it additionally exposes `log` and `sqrt` functions (which are used in `next_gaussian`). PR: https://github.com/loyispa/rust-strictmath/pull/3

muzarski:
v4:
- fetching `rust-strictmath` directly from `crates.io` as PR exposing `log` and `sqrt` functions has been merged to upstream.

piodul (approved):

Issue #52, opened 2023-09-15
Use a library to parse custom syntax in arguments


piodul:
The c-s and s-b frontends have some parameters with custom syntax. Currently, they are all parsed using ad-hoc routines that manipulate on strings, which are tedious to write and hard to read. We should consider pulling in a parsing library and defining the syntaxes of the parameters in a declarative way.

When choosing the library, we should optimize for code readability and clear error messages. The performance is completely irrelevant as parsing only happens at the program start.

Migrated to Jira: [QATOOLS-73](https://scylladb.atlassian.net/browse/QATOOLS-73)

[QATOOLS-73]: https://scylladb.atlassian.net/browse/QATOOLS-73?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #63, opened 2023-11-28
c-s mixed operation

muzarski:
# Motivation
Cassandra-stress supports mixed workloads. It allows to run multiple kinds of operations. Allowed operations are:
- read
- write
- counter_read
- counter_write

The parameters supported by `mixed` command are:
- all of the common parameters (e.g. used by `read`)
- all of the `counter_write` parameters
- `ratio()` parameter based on which tool samples an operation to execute. E.g. `ratio(read=2, write=1)` means that there will be approximately 2 read operations per 1 write operation.
- `clustering` distribution parameter which tells the tool how many times to run the operation that just got sampled.

# Changes
- Prepared `settings` module for `MixedOperation`
- Implemented `MixedOperation`

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
+++ b/src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
@@ -1,0 +1,33 @@
+use anyhow::Result;
+use rand_distr::{Distribution, WeightedIndex};
+
+#[derive(Clone)]
+pub struct EnumeratedDistribution<T> {
+    items: Vec<(T, f64)>,
+    dist: WeightedIndex<f64>,
+}
+
+impl<T: Copy> EnumeratedDistribution<T> {
+    pub fn new(items: Vec<(T, f64)>) -> Result<Self> {
+        let dist = WeightedIndex::new(items.iter().map(|w| w.1))?;
+
+        Ok(Self { items, dist })
+    }
+
+    pub fn sample(&self) -> T {
+        self.items[self.dist.sample(&mut rand::thread_rng())].0
+    }
+}
+
+impl<T: std::fmt::Display> std::fmt::Display for EnumeratedDistribution<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{{")?;
+        let items_str = self
+            .items
+            .iter()
+            .map(|item| format!("{}={}", item.0, item.1))
+            .collect::<Vec<_>>()
+            .join(",");
+        write!(f, "{}}}", items_str)
+    }
+}
--- a/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
@@ -1,0 +1,169 @@
+use anyhow::Result;
+use std::{ops::ControlFlow, sync::Arc};
+
+use cql_stress::{
+    configuration::{Operation, OperationContext, OperationFactory},
+    make_runnable,
+};
+use scylla::{frame::response::result::CqlValue, Session};
+
+use crate::{
+    java_generate::distribution::Distribution,
+    settings::{CassandraStressSettings, MixedSubcommand, OperationRatio},
+    stats::ShardedStats,
+};
+
+use super::{
+    counter_write::{CounterWriteOperation, CounterWriteOperationFactory},
+    read::{
+        CounterReadOperation, CounterReadOperationFactory, RegularReadOperation,
+        RegularReadOperationFactory,
+    },
+    row_generator::RowGenerator,
+    write::{WriteOperation, WriteOperationFactory},
+    CassandraStressOperation, CassandraStressOperationFactory, RowGeneratorFactory,
+    DEFAULT_COUNTER_TABLE_NAME, DEFAULT_TABLE_NAME,
+};
+
+pub struct MixedOperation {
+    write_operation: WriteOperation,
+    counter_write_operation: CounterWriteOperation,
+    read_operation: RegularReadOperation,
+    counter_read_operation: CounterReadOperation,
+    cached_row: Option<Vec<CqlValue>>,
+    workload: RowGenerator,
+    max_operations: Option<u64>,
+    stats: Arc<ShardedStats>,
+    operation_ratio: Arc<OperationRatio>,
+    clustering_distribution: Box<dyn Distribution>,
+    current_operation: MixedSubcommand,
+    current_operation_remaining: usize,
+}
+
+pub struct MixedOperationFactory {
+    settings: Arc<CassandraStressSettings>,
+    write_operation_factory: WriteOperationFactory,
+    counter_write_operation_factory: CounterWriteOperationFactory,
+    read_operation_factory: RegularReadOperationFactory,
+    counter_read_operation_factory: CounterReadOperationFactory,
+    operation_ratio: Arc<OperationRatio>,
+    workload_factory: RowGeneratorFactory,
+    max_operations: Option<u64>,
+    stats: Arc<ShardedStats>,
+}
+
+impl OperationFactory for MixedOperationFactory {
+    fn create(&self) -> Box<dyn Operation> {
+        let mixed_params = self.settings.command_params.mixed.as_ref().unwrap();
+
+        let write_operation = self.write_operation_factory.create();
+        let counter_write_operation = self.counter_write_operation_factory.create();
+        let read_operation = self.read_operation_factory.create();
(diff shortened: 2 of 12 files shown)


src/bin/cql-stress-cassandra-stress/settings/command/mixed.rs
```diff
@@ -0,0 +1,94 @@
...
+                    command
+                );
+                command_set.insert(command);
+                Ok((command, weight))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Self::new(&weights)
+    }
+}
```
piodul, line 89: Side note: this falls under the "custom syntax" mentioned in https://github.com/scylladb/cql-stress/issues/52 (no need to do anything right now).

src/bin/cql-stress-cassandra-stress/settings/command/mixed.rs
```diff
@@ -0,0 +1,94 @@
...
+            MixedSubcommand::CounterRead => "counter_read",
+            MixedSubcommand::CounterWrite => "counter_write",
+        };
+        write!(f, "{}", s)
+    }
+}
+
+pub type OperationRatio = EnumeratedDistribution<MixedSubcommand>;
+
+impl OperationRatio {
```
piodul, line 37: Please include a short comment which describes the syntax.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
```diff
@@ -0,0 +1,36 @@
...
+use rand_distr::{Distribution, WeightedIndex};
+
+#[derive(Clone)]
+pub struct EnumeratedDistribution<T: Copy + std::fmt::Display> {
+    items: Vec<(T, f64)>,
+    dist: WeightedIndex<f64>,
+}
+
+impl<T: Copy + std::fmt::Display> EnumeratedDistribution<T> {
+    pub fn new(items: &[(T, f64)]) -> Result<Self> {
```
piodul, line 11: Nit: take `items` by value (i.e. `items: Vec<(T, f64)>`). If the caller prepared a `Vec<(T, f64)>` and no longer uses it, it will let them avoid a copy. In other cases, there should be no problem for the caller copy.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
```diff
@@ -0,0 +1,36 @@
+use anyhow::Result;
+use rand_distr::{Distribution, WeightedIndex};
+
+#[derive(Clone)]
+pub struct EnumeratedDistribution<T: Copy + std::fmt::Display> {
```
piodul, line 5: As a principle, trait bounds should be avoided as possible in struct/enum definitions, unless they are required for the struct/enum definition to compile. https://stackoverflow.com/a/66369912

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -0,0 +1,181 @@
...
+        let mixed_params = settings.command_params.mixed.as_ref().unwrap();
+        let read_statement =
+            prepare_read_statement(regular_table_name, &session, &settings).await?;
+        let counter_read_statement =
+            prepare_read_statement(counter_table_name, &session, &settings).await?;
+        let write_statement = WriteOperationFactory::prepare_statement(&session, &settings).await?;
+        let counter_write_statement =
+            CounterWriteOperationFactory::prepare_statement(&session, &settings).await?;
+        let max_operations = settings.command_params.common.operation_count;
+        let operation_ratio = Arc::new(mixed_params.operation_ratio.clone());
```
piodul, line 101: I'm not a fan of the approach... You are exposing some details of the existing operations and use them to implement `MixedOperation`.

I think that the operations should be self-contained. `MixedOperation` should just encapsulate them and not care about their internals apart from the fact that they implement some trait. Would it be possible to implement it like that?

piodul (changes requested):

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
muzarski: I thought about introducing some trait, but don't we end up with the same performance issue as in https://github.com/scylladb/cql-stress/pull/34 (`async-trait` boxed futures)?

I haven't compared both of the approaches when it comes to performance, but to me, it looks exactly the same - we end up with per-operation allocation from boxing the futures. I can still do the comparison, if necessary.

piodul: I see. AFAIK `async` in trait methods is going to be stabilized in Rust 1.75, so maybe we can use `async_trait` until that happens.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
```diff
@@ -0,0 +1,36 @@
...
+
+        Ok(Self {
+            items: items.to_vec(),
+            dist,
+        })
+    }
+
+    pub fn sample(&self) -> T {
+        self.items[self.dist.sample(&mut rand::thread_rng())].0
+    }
```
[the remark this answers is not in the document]
muzarski: We do not depend on any determinism here, so we don't need to use a `java_random::Random`.

In addition, `rand_distr::WeightedIndex::sample` accepts `&mut R` where `R: rand::Rng`. We would need to implement `rand::RngCore` for `java_random::Random` (which is impossible, so we would need to wrap it with yet another type).

I just decided to keep it simple and use the rng provided by `rand` crate.

roydahan:
@muzarski let's try to move this one forward, once we will have this option supported we wlll be able to replace some of the currentl longevities that are using cassandra-stress with cql-stress.

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
muzarski: > I see. AFAIK `async` in trait methods is going to be stabilized in Rust 1.75, so maybe we can use `async_trait` until that happens.

They indeed got stabilized, but unfortunately such traits are not object-safe, and so cannot be used in trait objects. I think the issue still holds. Do you still think we should use `async_trait`?

muzarski:
v2:
- addressed review comments
- rebased on top of https://github.com/scylladb/cql-stress/pull/68. Can be useful to review mentioned PR to see how the changes interact with `mixed` command.

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -0,0 +1,169 @@
...
+                        .generate_row(&mut self.workload)
+                });
+                self.counter_write_operation.execute(row).await
+            }
+        };
+
+        self.stats.get_shard_mut().account_operation(ctx, &result);
+
+        if result.is_ok() {
+            self.current_operation_count = self.current_operation_count.saturating_sub(1);
```
piodul, line 163: Is there a particular reason here for using saturating_sub? We check for `self.current_operation_count == 0` earlier in the function, so the current operation count here should not be `0`, right?

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -0,0 +1,169 @@
...
+            read_operation,
+            counter_read_operation,
+            cached_row: None,
+            workload: self.workload_factory.create(),
+            max_operations: self.max_operations,
+            stats: Arc::clone(&self.stats),
+            operation_ratio: Arc::clone(&self.operation_ratio),
+            clustering_distribution: mixed_params.clustering.create(),
+            current_operation: MixedSubcommand::Read,
+            current_operation_count: 0,
```
piodul, line 76: How about `current_operation_remaining`? It will be more descriptive IMO.

piodul (commented):
Left some nitpicks, after fixing them and rebasing it will LGTM.

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
muzarski: Yes, it's pointless to have saturating_sub here. It remains from the previous version where we would have:
```rust
self.current_operation_count = self.clustering_distribution.next_i64() as usize;
```
instead of
```rust
self.current_operation_count =
    (self.clustering_distribution.next_i64() as usize).max(1);
```

In the previous version there was a possibility that `self.current_operation_count` was 0 after taking the sample. Now we are guaranteed that it's at least 1 (thanks to `usize::max` method).

I'll get rid of the `saturating_sub`.

piodul (approved):

Issue #67, opened 2024-01-19
c-s operation: retry operation with the same partition key


muzarski:
As of now, there is a bug in the c-s frontend logic where we generate a new partition key each time we retry an operation.

It needs to be fixed so we cache the sampled partition key and reuse it during retries.

PR #68, opened 2024-02-21
c-s operation: Extract common operation logic

muzarski:
Fixes: https://github.com/scylladb/cql-stress/issues/67

## Motivation

Currently, the common logic for operations is repeated. The idea is to introduce an additional struct that will handle the common logic and encapsulate the logic of specific c-s operation.

This will also help to implement `mixed` operation. This is why, we should merge it before https://github.com/scylladb/cql-stress/pull/63.

## Solution

Introduced a `CassandraStressOperation` trait which represents a specific c-s operation.
The `GenericCassandraStressOperation` structure will handle the common logic and delegate
the specific operation logic to some object implementing `CassandraStressOperation`.

The common logic that was extracted to `GenericCassandraStressOperation`:
- checks whether `operation_id` > `max_operations`. If true, the stress test will be terminated
- recording operation result to stats structure
- caching the row for retries (https://github.com/scylladb/cql-stress/issues/67)

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/operation/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mod.rs
@@ -14,7 +14,6 @@ use std::num::Wrapping;
 use std::ops::ControlFlow;
 use std::sync::Arc;
 
-pub use counter_write::CounterWriteOperationFactory;
 pub use row_generator::RowGeneratorFactory;
 use scylla::{
     frame::response::result::{CqlValue, Row},
@@ -117,6 +116,8 @@ pub struct GenericCassandraStressOperationFactory<O: CassandraStressOperation> {
 }
 
 pub type WriteOperationFactory = GenericCassandraStressOperationFactory<write::WriteOperation>;
+pub type CounterWriteOperationFactory =
+    GenericCassandraStressOperationFactory<counter_write::CounterWriteOperation>;
 pub type RegularReadOperationFactory =
     GenericCassandraStressOperationFactory<read::RegularReadOperation>;
 pub type CounterReadOperationFactory =
@@ -131,6 +132,26 @@ impl WriteOperationFactory {
     ) -> Result<Self> {
         let max_operations = settings.command_params.common.operation_count;
         let cs_operation_factory = write::WriteOperationFactory::new(settings, session).await?;
+
+        Ok(Self {
+            cs_operation_factory,
+            max_operations,
+            workload_factory,
+            stats,
+        })
+    }
+}
+
+impl CounterWriteOperationFactory {
+    pub async fn new(
+        settings: Arc<CassandraStressSettings>,
+        session: Arc<Session>,
+        workload_factory: RowGeneratorFactory,
+        stats: Arc<ShardedStats>,
+    ) -> Result<Self> {
+        let max_operations = settings.command_params.common.operation_count;
+        let cs_operation_factory =
+            counter_write::CounterWriteOperationFactory::new(settings, session).await?;
 
         Ok(Self {
             cs_operation_factory,
--- a/src/bin/cql-stress-cassandra-stress/operation/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mod.rs
@@ -4,7 +4,15 @@ mod row_generator;
 mod write;
 
 use anyhow::Result;
+use cql_stress::configuration::Operation;
+use cql_stress::configuration::OperationContext;
+use cql_stress::configuration::OperationFactory;
+use cql_stress::make_runnable;
+use scylla::Session;
+use std::future::Future;
 use std::num::Wrapping;
+use std::ops::ControlFlow;
+use std::sync::Arc;
 
 pub use counter_write::CounterWriteOperationFactory;
 pub use read::CounterReadOperationFactory;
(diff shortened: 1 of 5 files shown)


src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -16,6 +23,121 @@ use scylla::{
...
+
+use self::row_generator::RowGenerator;
+
+/// A specific CassandraStress operation.
+#[trait_variant::make(CassandraStressOperation: Send)]
+pub trait LocalCassandraStressOperation: Sync + Send + Sized {
+    type Factory: CassandraStressOperationFactory<Operation = Self>;
+
+    const DEFAULT_TABLE_NAME: &'static str = "standard1";
+    const DEFAULT_COUNTER_TABLE_NAME: &'static str = "counter1";
```
piodul, line 33: Not sure that they should be associated constants. Are they supposed to be overriden? If not then I suggest making them module-level constants.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -16,6 +23,121 @@ use scylla::{
...
+
+    fn get_stressed_table_name() -> &'static str;
+    async fn execute(&self, row: &[CqlValue]) -> Result<ControlFlow<()>>;
```
piodul, line 36: Please document how this trait interacts with `GenericCassandraStressOperation` - most importantly, how `CassandraStressOperation::execute`'s return value should be interpreted.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -16,6 +23,121 @@ use scylla::{
...
+
+/// A specific CassandraStress operation.
+#[trait_variant::make(CassandraStressOperation: Send)]
+pub trait LocalCassandraStressOperation: Sync + Send + Sized {
+    type Factory: CassandraStressOperationFactory<Operation = Self>;
+
+    const DEFAULT_TABLE_NAME: &'static str = "standard1";
+    const DEFAULT_COUNTER_TABLE_NAME: &'static str = "counter1";
+
+    fn get_stressed_table_name() -> &'static str;
```
piodul, line 35: At a glance, it looks like this method is called only from within other methods from the same implementation of the trait. Perhaps we can drop it?

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -16,6 +23,121 @@ use scylla::{
...
 pub use write::WriteOperationFactory;
 
+use crate::settings::CassandraStressSettings;
+use crate::stats::ShardedStats;
+
+use self::row_generator::RowGenerator;
+
+/// A specific CassandraStress operation.
+#[trait_variant::make(CassandraStressOperation: Send)]
+pub trait LocalCassandraStressOperation: Sync + Send + Sized {
```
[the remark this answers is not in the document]
muzarski: > Why `LocalCassandraStressOperation` requires `Sync` and `Send`? Shouldn't it require neither?

You are right, these should be moved to non-local version. I forgot to move them after introducing `trait-variant`.

> I'm not sure why a local variant is needed at all.

It actually is not, I just decided to import the `trait-variant` crate and make use of the macro instead of putting the trait bounds on the return future manually (`impl Future<Output = Result<ControlFlow<()>> + Send + Sync>`). I thought that this may look better.

WDYT?

src/bin/cql-stress-cassandra-stress/operation/mod.rs
muzarski: Actually, this function is only used in implementations of `CassandraStressOperationFactory` during creation of `PreparedStatement`.

It is mainly needed to differentiate between `read` and `counter_read` operations (see `read.rs`). One of them reads from `standard1` table while the other reads from `counter1` table.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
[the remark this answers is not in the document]
piodul: > It actually is not, I just decided to import the `trait-variant` crate and make use of the macro instead of putting the trait bounds on the return future manually (`impl Future<Output = Result<ControlFlow<()>> + Send + Sync>`). I thought that this may look better.
> 
> WDYT?

I think it's confusing to use `trait-variant` for that. AFAIK it's primary purpose was to make it possible to quickly generate variants of the trait with different bounds. Using it here implies that the local variant of the trait has some purpose, while it is not used anywhere.

There is only one `async` method in the trait so I think it won't be too bad if we use return position impl in trait (RPITIT) here.

As a side note, why doesn't `CassandraStressOperationFactory::create` have this problem?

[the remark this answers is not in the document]
muzarski: So I'll remove the `trait-variant` dependency at all for now.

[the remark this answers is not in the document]
muzarski: @piodul I think that the `Send` bound actually comes from the `async_trait`, not from the `tokio::task::spawn`. The `async_trait` macro on `src/configuration.rs::Operation` expands to:
```rust
fn run<'life0, 'async_trait>(
        &'life0 mut self,
        session: WorkerSession,
    ) -> ::core::pin::Pin<
        Box<dyn ::core::future::Future<Output = Result<()>> + ::core::marker::Send + 'async_trait>,
    >
    where
        'life0: 'async_trait,
        Self: 'async_trait;
```

The error message from the compiler is not clear to me and that's why I assumed that it comes from the `tokio::task::spawn` bound.

For reference - the error message when I remove the `Send` bound from `CassandraStressOperation::execute`:
```
error: future cannot be sent between threads safely
  --> src/bin/cql-stress-cassandra-stress/operation/mod.rs:66:1
   |
66 | make_runnable!(GenericCassandraStressOperation<O: CassandraStressOperation>);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
   |
   = help: within `{async block@/home/mikolajuzarski/proj/cql-stress/src/configuration.rs:117:100: 125:14}`, the trait `Send` is not implemented for `impl Future<Output = std::result::Result<ControlFlow<()>, anyhow::Error>>`
note: future is not `Send` as it awaits another future which is not `Send`
  --> src/bin/cql-stress-cassandra-stress/operation/mod.rs:80:25
   |
80 |         let op_result = self.cs_operation.execute(row).await;
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ await occurs here on type `impl Future<Output = std::result::Result<ControlFlow<()>, anyhow::Error>>`, which is not `Send`
   = note: required for the cast from `Pin<Box<{async block@/home/mikolajuzarski/proj/cql-stress/src/configuration.rs:117:100: 125:14}>>` to `Pin<Box<dyn Future<Output = std::result::Result<(), anyhow::Error>> + Send>>`
   = note: this error originates in the macro `make_runnable` (in Nightly builds, run with -Z macro-backtrace for more info)

```

src/bin/cql-stress-cassandra-stress/operation/mod.rs
piodul: In that case, there is a bug in `read.rs`, in `GenericReadOperationFactory::new`, it uses a hardcoded name of `"standard1"` to create the prepared statement.

Still, I'm not sure I like the idea to put this into `CassandraStressOperation`. It's only really used in the case you mentioned; although there are actually two other cases in `counter_write.rs` and `write.rs`, I don't think they count because they spell the implementing type directly and we don't really utilize the fact that this is a trait method there.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -16,6 +23,121 @@ use scylla::{
...
+
+    fn get_stressed_table_name() -> &'static str;
+    async fn execute(&self, row: &[CqlValue]) -> Result<ControlFlow<()>>;
+    fn generate_row(&self, row_generator: &mut RowGenerator) -> Vec<CqlValue>;
+}
+
+pub trait CassandraStressOperationFactory: Sync + Send + Sized {
+    type Operation: CassandraStressOperation<Factory = Self>;
+
+    async fn new(settings: Arc<CassandraStressSettings>, session: Arc<Session>) -> Result<Self>;
```
piodul, line 43: Do we actually need this to be a trait method? I think that `GenericCassandraStressOperationFactory` could just have a `new` method which takes something that implements `CassandraStressOperationFactory`. I don't think it will inflate code much, and I suppose that it will make it easier to get rid of `get_stressed_table_name` - for the read operation factory, you could have struct field that defines the table name.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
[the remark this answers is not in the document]
piodul: Makes sense, `async_trait` probably tries to make the returned futures useful in tokio tasks (or, in general, with work stealing executors), so it requires `Send`.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
muzarski: Yep, I see your point - makes sense to me.

muzarski:
v2: addressed review comments

piodul (approved):

PR #69, opened 2024-02-22
c-s: `-mode` option

muzarski:
# Motivation
Cassandra-stress supports `-mode` CLI option which configures the connection with database.

# Changes

In this PR we introduce only a part of CLI parameters supported by `-mode` option, which are broadly used in SCT:
- `compression=` - specify a compression algorithm used for connections. Can be one of: `lz4`, `snappy` or `none` (no compression at all)
- `user=`, `password=` - they both specify user credentials for connections
- `connectionsPerHost=`, `connectionsPerShard=` - the latter one is a bonus parameter introduced in cql-stress - previously introduced in https://github.com/scylladb/cql-stress/pull/55 as a `-node` option parameter. These two parameters specify the number of opened connections per host and per shard respectively.
- `cql3`, `native` - dummy parameters used in SCT. Cassandra-Stress accepts `simplenative` and `thrift` modes which are obviously not supported by Rust driver. These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.

I decided not to introduce following parameters - I don't think they are any useful + they are not used in SCT at all:
- `unprepared` - force use of unprepared statements
- `auth-provider=?` - Fully qualified implementation of com.datastax.driver.core.AuthProvider
- `maxPending=?` - Maximum pending requests per connection
- `protocolVersion=?` - CQL Protocol Version
- `port=?` - No description provided by cassandra-stress (?). From what I've seen in the code, the parameter is not used at all, just parsed. The default value is 9046 (any ideas?).


fruch (approved):

piodul:
> * `cql3`, `native` - dummy parameters used in SCT. Cassandra-Stress accepts `simplenative` and `thrift` modes which are obviously not supported by Rust driver. These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.

I'm not sure we should plainly ignore them. If we have some tests in SCT which use thrift then I think that cql-stress should return an error if one tries to pass this option - so that a test does not accidentally run with the CQL protocol.

piodul:
> * `connectionsPerHost=`, `connectionsPerShard=` - the latter one is a bonus parameter introduced in cql-stress - previously introduced in

How does the original c-s with our fork of the driver treat `connectionsPerHost`, does it work as one would imagine or is it ignored? Do we use this parameter in SCT intentionally?

fruch:
> > * `cql3`, `native` - dummy parameters used in SCT. Cassandra-Stress accepts `simplenative` and `thrift` modes which are obviously not supported by Rust driver. These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.
> 
> I'm not sure we should plainly ignore them. If we have some tests in SCT which use thrift then I think that cql-stress should return an error if one tries to pass this option - so that a test does not accidentally run with the CQL protocol.

we don't use thrift in SCT, it's deprecated for quite some time, I don't think it would ever be used.

fruch:
> > * `connectionsPerHost=`, `connectionsPerShard=` - the latter one is a bonus parameter introduced in cql-stress - previously introduced in
> 
> How does the original c-s with our fork of the driver treat `connectionsPerHost`, does it work as one would imagine or is it ignored? Do we use this parameter in SCT intentionally?

it works as describe, and we use it in SCT, in two test cases, in one of them to generate lots many connection as possible.

src/bin/cql-stress-cassandra-stress/settings/option/mode.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+fn prepare_parser() -> (ParamsParser, ModeParamHandles) {
+    let mut parser = ParamsParser::new(ModeOption::CLI_STRING);
+
+    // `cql3` and `native` parameters are ignored but we introduce them so we support
+    // the cassandra-stress CLI syntax which is used in SCT.
+    let _cql3 = parser.simple_param::<bool>("cql3", None, "cql3 mode (dummy parameter)", false);
+    let _native =
+        parser.simple_param::<bool>("native", None, "native mode (dummy parameter)", false);
```
piodul, line 88: Why do those variable names start with underscores? Usually it is done for variables that are not used, but they are used in this function.

piodul: While I get it that the _parameters_ are not used in the program, the variables themselves are used.

src/bin/cql-stress-cassandra-stress/settings/option/mode.rs
```diff
@@ -0,0 +1,212 @@
...
+        "connectionsPerHost=",
+        None,
+        "Number of connections per host",
+        false,
+    );
+
+    // $ ./cql-stress-cassandra-stress help -node
+    // Usage: -mode cql3 native [compression=?] [user=?] [password=?] [connectionsPerShard=?]
+    //  OR
+    // Usage: -mode cql3 native [compression=?] [user=?] [password=?] [connectionsPerHost=?]
```
piodul, line 114: If we add more mutually exclusive groups of parameters like `connectionsPerHost` + `connectionsPerShar` in the future, we might get a combinatorial explosion with this approach.

Just pointing that out as a potential problem in the future, let's not worry about it too much right now.

piodul:
> > > * `cql3`, `native` - dummy parameters used in SCT. Cassandra-Stress accepts `simplenative` and `thrift` modes which are obviously not supported by Rust driver. These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.
> > 
> > 
> > I'm not sure we should plainly ignore them. If we have some tests in SCT which use thrift then I think that cql-stress should return an error if one tries to pass this option - so that a test does not accidentally run with the CQL protocol.
> 
> we don't use thrift in SCT, it's deprecated for quite some time, I don't think it would ever be used.

If that is the case, then it shouldn't be a problem to return an error when `thrift` is provided. Maybe we can skip adding this parameter altogether.

In fact, I checked out the code of the PR and neither `thrift` nor `simplenative` appears anywhere in the code. So it looks like it's not that we parse the parameters but ignore them, but rather we decided to ignore the parameters by not implementing them. I think this is the right approach, but the wording in the PR description was confusing.

muzarski:
> > > > * `cql3`, `native` - dummy parameters used in SCT. Cassandra-Stress accepts `simplenative` and `thrift` modes which are obviously not supported by Rust driver. These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.
> > > 
> > > 
> > > I'm not sure we should plainly ignore them. If we have some tests in SCT which use thrift then I think that cql-stress should return an error if one tries to pass this option - so that a test does not accidentally run with the CQL protocol.
> > 
> > 
> > we don't use thrift in SCT, it's deprecated for quite some time, I don't think it would ever be used.
> 
> If that is the case, then it shouldn't be a problem to return an error when `thrift` is provided. Maybe we can skip adding this parameter altogether.
> 
> In fact, I checked out the code of the PR and neither `thrift` nor `simplenative` appears anywhere in the code. So it looks like it's not that we parse the parameters but ignore them, but rather we decided to ignore the parameters by not implementing them. I think this is the right approach, but the wording in the PR description was confusing.

Yes, sorry for the confusion. The part: `These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.` of the description relate to `cql3` and `native` parameters which are ignored - not the `thrift` and `simplenative`.

src/bin/cql-stress-cassandra-stress/settings/option/mode.rs
muzarski: Yeah, I agree - I think the parameters' grouping should be somehow refactored in the future. As of now it's not very convenient to work with the parameter groups but at least it works..

src/bin/cql-stress-cassandra-stress/settings/option/mode.rs
muzarski: I wanted to emphasize the fact that these parameters are ignored and not used by the runtime. But I see your point and I understand the intent behind the names starting with underscore - I'll remove the underscores then.

piodul:
> Yes, sorry for the confusion. The part: `These two parameters are simply ignored but introduced so cassandra-stress commands can be easily replaced with cql-stress commands in SCT.` of the description relate to `cql3` and `native` parameters which are ignored - not the `thrift` and `simplenative`.

Then please rephrase the paragraph to be less ambiguous about this.

muzarski:
v2: rebased on top of master

muzarski:
v3: removed underscores from variable names

src/bin/cql-stress-cassandra-stress/settings/option/mode.rs
```diff
@@ -0,0 +1,212 @@
...
+        ParamsParser, SimpleParamHandle,
+    },
+    ParsePayload,
+};
+
+pub struct ModeOption {
+    pub compression: Option<Compression>,
+    pub user_credentials: Option<(String, String)>,
+    pub pool_size: PoolSize,
+}
```
wprzytula, line 16: Instead of using an unnamed tuple for credentials and relying on implicit ordering of it, I suggest creating a dedicated struct for them here.

wprzytula:
```suggestion

struct Credentials {
   username: String,
   password: String,
}

pub struct ModeOption {
    pub compression: Option<Compression>,
    pub user_credentials: Option<Credentials>,
    pub pool_size: PoolSize,
}
```

muzarski:
v4: addressed review comment

muzarski:
@piodul can we merge this one?

wprzytula (approved):

piodul (approved):

PR #71, opened 2024-03-08
Bump driver version to 0.12.0

muzarski:
# Changes
Bumped driver's version to most recent `0.12.0`.

Some code needed to be adjusted:
## `Consistency` enum
`Consistency` enum supports two new variants: `Serial` and `LocalSerial`. Updated `show_consistency_level` accordingly.

## `Murmur3Partitioner`
`Murmur3Partitioner`'s API changed a bit, so its usage is adjusted in `GeneratorConfig::new`

## new serialization api (0.11.0)
`SerializedValues` was renamed to `LegacySerializedValues` in serialization API refactor.

I changed the usage of `SerializedValues` simply to `Vec<i64>` in s-b's `ReadOperation`. I used `Vec<i64>` since both `partition_key` and `clustering_key`s are `bigint`s.

Maybe it's better use `LegacySerializedValues`, but I'm not quite sure as I'm not familiar with the changes introduced in 0.11.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Lorak-mmk:
> ## new serialization api (0.11.0)
> 
> `SerializedValues` was renamed to `LegacySerializedValues` in serialization API refactor.
> 
> I changed the usage of `SerializedValues` simply to `Vec<i64>` in s-b's `ReadOperation`. I used `Vec<i64>` since both `partition_key` and `clustering_key`s are `bigint`s.
> 
> Maybe it's better use `LegacySerializedValues`, but I'm not quite sure as I'm not familiar with the changes introduced in 0.11.0.


You should not use legacy structs, they will be removed in the future.
Current approach (`Vec<i64>`) is ok of all your values are i64.

Lorak-mmk (approved):

piodul (approved):

Issue #73, opened 2024-03-27
incompatible CLI with cassandra-stress


soyacz:
cql-stress-cassandra-stress uses different CLI arguments for `-col` argument. To make it work need to switch `-col 'size=FIXED(128) n=FIXED(8)'` to  `-col 'size=FIXED(128) n=8'`
also for `-pop` arg it is used differently: `-pop seq=1..125000000` need to switch to `-pop 'dist=SEQ(1..125000000)'` in cql-stress

This makes replacement of c-s to cql-stress more cumbersome and leads to failures.

avelanarius:
> also for -pop arg it is used differently: -pop seq=1..125000000 need to switch to -pop 'dist=SEQ(1..125000000)' in cql-stress

In the current master version of cql-stress this is already solved.

avelanarius:
> To make it work need to switch -col 'size=FIXED(128) n=FIXED(8)' to -col 'size=FIXED(128) n=8'

... but this still is not fixed.

PR #75, opened 2024-04-05
c-s/col: c-s compatibility for -col n= parameter

muzarski:
Fixes: https://github.com/scylladb/cql-stress/issues/73

cassandra-stress supports a distribution syntax for `-col n=` parameter. However, it only allows the `FIXED(?)` distribution for cql mode.

Before this commit, cql-stress would accept a u64 value for the value of this parameter.

To make our CLI compatible with c-s (so it can be easily replaced in SCT), we will accept `FIXED(<u64>)` syntax as well.

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/settings/option/column.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/option/column.rs
@@ -1,9 +1,13 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
+use cql_stress::distribution::{parse_description, SyntaxFlavor};
 
 use crate::{
     java_generate::distribution::DistributionFactory,
     settings::{
-        param::{types::CommaDelimitedList, ParamsParser, SimpleParamHandle},
+        param::{
+            types::{CommaDelimitedList, Parsable},
+            ParamsParser, SimpleParamHandle,
+        },
         ParsePayload,
     },
 };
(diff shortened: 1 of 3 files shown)


src/bin/cql-stress-cassandra-stress/settings/option/column.rs
```diff
@@ -57,9 +61,41 @@ impl ColumnOption {
...
+
+        || -> Result<u64, anyhow::Error> {
+            let s = &s.to_lowercase();
+            let description = parse_description(s, SyntaxFlavor::Classic)
+                .context("Failed to parse distribution description.")?;
+            anyhow::ensure!(description.name == "fixed", "Expected FIXED distribution.");
+            description.check_argument_count(1)?;
+            u64::parse(description.args[0]).context("Failed to parse u64 value")
+        }()
+        .context("Invalid value. Available values are either <u64> or FIXED(<u64>).")
```
soyacz, line 92: will error message contain info that this error refers to column count?

soyacz (approved):
Besides small question, LGTM

src/bin/cql-stress-cassandra-stress/settings/option/column.rs
muzarski: yes, it's already handled by other piece of code

src/bin/cql-stress-cassandra-stress/settings/option/column.rs
```diff
@@ -57,9 +61,41 @@ impl ColumnOption {
...
+        }
+
+        || -> Result<u64, anyhow::Error> {
+            let s = &s.to_lowercase();
+            let description = parse_description(s, SyntaxFlavor::Classic)
+                .context("Failed to parse distribution description.")?;
+            anyhow::ensure!(description.name == "fixed", "Expected FIXED distribution.");
+            description.check_argument_count(1)?;
+            u64::parse(description.args[0]).context("Failed to parse u64 value")
+        }()
```
wprzytula, line 91: What a nice trick for seamless use of `?` operator - immediately called closure. Have you been using this elsewhere as well yet?

src/bin/cql-stress-cassandra-stress/settings/option/column.rs
```diff
@@ -57,9 +61,41 @@ impl ColumnOption {
     }
 }
 
+/// A type for parsing `-col n=` parameter.
+///
+/// In cassandra-stress, CLI originally accepts a distribution.
+/// However, only FIXED(?) distribution is supported by the tool.
```
wprzytula, line 67: It is unclear to me whether `the tool` refers to cassandra-stress or cql-stress.

src/bin/cql-stress-cassandra-stress/settings/option/column.rs
muzarski: I think I've used the trick once or twice before in cql-stress in very similar context - I wanted to attach some common error message at the end of the function in case of failure. This way we don't have to introduce an additional function/method or variable (closure which has to be then called).

wprzytula (approved):

PR #77, opened 2024-05-08
readme: c-s frontend description

muzarski:
depends on https://github.com/scylladb/cql-stress/pull/74

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,11 @@ The `scylla-bench` frontend is feature-complete, `cassandra-stress` is a _work i
 `cql-stress` is not published on crates.io yet, therefore in order to use it you need to clone it and build from source.
 See the [Development](#development) section for more details.
 
+## Crate features
+
+List of the crate features:
+- `user-profile` - enables support for `user` command and custom user profiles in `cassandra-stress` frontend. This feature is enabled by default. To disable it, pass `--no-default-features` flag when building the tool.
+
 ### Scylla Bench
 
 See the documentation of the original [`scylla-bench`](https://github.com/scylladb/scylla-bench/blob/master/README.md#usage) for a comprehensive explanation of the most important parameters.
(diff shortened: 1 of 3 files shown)


soyacz:
can you verify with user profiles in https://github.com/scylladb/scylla-qa-internal/tree/master repo if c-s frontend works with them?

muzarski:
CI fails due to clippy errors, see https://github.com/scylladb/cql-stress/pull/81

README.md
```diff
@@ -17,7 +17,36 @@ To see a list of all the parameters currently supported by the tool, use `cql-st
...
+To see a list of all commands and options currently supported by the tool, use `cql-stress-cassandra-stress help`. To see a list of parameters supported for a given command/option, use `cql-stress-cassandra-stress help <command/option>`.
+
+#### Populating the cluster
+
+To populate a local cluster, make use of `write` command:
+```
+cql-stress-cassandra-stress write n=1000000 -pop seq=1..1000000 -rate threads=20 -node 127.0.0.1
+```
+
+Since some of the options and paramters were not provided, the tool will make use of some default values. This will result in:
```
wprzytula, line 31: parameters

README.md
```diff
@@ -17,7 +17,36 @@ To see a list of all the parameters currently supported by the tool, use `cql-st
...
+cql-stress-cassandra-stress read n=1000000 -pop seq=1..1000000 -rate threads=20 -node 127.0.0.1
+```
+
+#### User profiles
+
+Commands mentioned above are very limited. They do not, for example, allow to test other native types than `blob`.
+
+To test more complex schemas, make use of user profiles (`user` command). User profiles allow to define custom schemas and custom statements used to stress the database.
+
+To enable the `user` mode, the tool needs to be compiled with `user-profile` feature: `cargo build --release --features "user-profile"`
```
wprzytula, line 49: As @avelanarius mentioned, probably most of c-s users are using user profiles. Perhaps it makes sense then to make the feature turned on by default, so less fuss is needed for them.

wprzytula (changes requested):

muzarski:
v2:
- enabled `user-profile` feature by default
- adjusted CI
- added mention about available crate features in README

wprzytula (approved):

muzarski:
> can you verify with user profiles in https://github.com/scylladb/scylla-qa-internal/tree/master repo if c-s frontend works with them?

I see that user profiles in this repo make use of `insert` and `columnspec` parameters. These won't unfortunately be supported in the very first version of user profiles in cql-stress. I think that for `columnspec`, it will be easy to introduce it, since we already have the logic implemented.

Issue #79, opened 2024-05-09
c-s: Implement runtime for `user` command


muzarski:
When https://github.com/scylladb/cql-stress/pull/74 is merged, the runtime for `user` command needs to be implemented.

The first version will sample the operations provided in profile in a round-robin fashion.

Issue #80, opened 2024-05-09
c-s: Implement operation sampler


muzarski:
As mentioned in https://github.com/scylladb/cql-stress/issues/79, the first version of runtime will sample the operations in round-robin fashion. However, in the original cassandra-stress, users can provide `ops()` parameter which defines ratios for operations with which they should be sampled.

There are two things to be done:
- implement `ops()` parameter for CLI
- refactor the sampler already implemented for `mixed` operation (`ratio()` parameter). It should be made generic, so it can be easily reused for `user` command.

PR #82, opened 2024-05-22
c-s user: Implement runtime for user command

muzarski:
Fix: https://github.com/scylladb/cql-stress/issues/79

This PR introduces a simple runtime for `user` command.

One operation consists of:
- sampling a pair (row, statement to execute). For now, the statements (provided by the user via `queries` parameter in yaml profile) are sampled in a round-robin fashion but this will be randomized in the future. See the comment in the code next to `OperationSampler`. Related issue: https://github.com/scylladb/cql-stress/issues/80
- Execute a sampled statement with variables bound from the sampled rows. If the operation fails, it will be retried with the same statement and row. This is why both of these are cached in case of failure.

## Tests

This PR introduces a test case that checks compatibility with original c-s for the `blob` CQL type. To see the tests for other types, see this PR https://github.com/scylladb/cql-stress/pull/83.


src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
```diff
@@ -143,7 +143,7 @@ impl RowGeneratorFactory {
             .iter()
             .map(|column| {
                 Generator::new(
-                    Blob::default(),
+                    Box::<Blob>::default(),
                     GeneratorConfig::new(
```
wprzytula, line 185: Nice idiom for initializing trait objects!

src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs
```diff
@@ -96,6 +96,11 @@ pub trait ValueGenerator: Send + Sync + 'static {
     ) -> CqlValue;
 }
 
+#[cfg(feature = "user-profile")]
+pub trait ValueGeneratorFactory: Send + Sync {
+    fn create(&self) -> Box<dyn ValueGenerator>;
+}
+
```
wprzytula, line 146: What's the purpose of this trait?
Please add docstrings.

wprzytula: Also, the commit message misses context (probably knowledge about c-s internals) that is likely necessary here.

src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs
```diff
@@ -40,6 +45,35 @@ impl Generator {
         }
     }
 
+    #[cfg(feature = "user-profile")]
+    pub fn new_generator_factory_from_cql_type(
+        type_: &CqlType,
```
wprzytula, line 54: personal nit: `typ` looks better, `type_` is just ugly.

src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs
```diff
@@ -40,6 +45,35 @@ impl Generator {
...
+                type_
+            ),
+            CqlType::Tuple(_) => anyhow::bail!(
+                "Invalid column type: {:?}. Tuples are not yet supported by the tool!",
+                type_
+            ),
+            CqlType::UserDefinedType { .. } => anyhow::bail!(
+                "Invalid column type: {:?}. UDTs are not yet supported by the tool!",
+                type_
+            ),
```
wprzytula, line 77: "Invalid" or rather "Unsupported" ?

src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
```diff
@@ -78,6 +80,41 @@ pub struct RowGenerator {
...
+    ) -> Self {
+        #[cfg(feature = "user-profile")]
+        let index_map = {
+            let mut index_map = HashMap::new();
+
+            index_map.insert(pk_generator.get_col_name().to_owned(), 0);
+            for (i, gen) in column_generators.iter().enumerate() {
+                index_map.insert(gen.get_col_name().to_owned(), i + 1);
+            }
+            index_map
```
wprzytula, line 102: WDYT about FP here?

wprzytula:
```suggestion
        let index_map = HashMap::from_iter(
            std::iter::once(&pk_generator).chain(column_generators.iter())
                .map(|generator| generator.get_col_name().to_owned())
                .enumerate()
                .map(|(i, col)| (col, i)
        );
```

src/bin/cql-stress-cassandra-stress/operation/row_generator.rs
```diff
@@ -78,6 +80,41 @@ pub struct RowGenerator {
...
+            column_generators,
+            #[cfg(feature = "user-profile")]
+            index_map,
+        }
+    }
+
+    #[cfg(feature = "user-profile")]
+    pub fn index_of(&self, name: &str) -> Option<usize> {
+        self.index_map.get(name).copied()
+    }
```
wprzytula, line 117: Perhaps a more explicit name? E.g. `statement_index_of_column_with_name`, but maybe you can come up with something more concise.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -227,7 +227,7 @@ fn recompute_seed(seed: i64, partition_key: &CqlValue) -> i64 {
         CqlValue::Blob(key) => {
             let mut wrapped = Wrapping(seed);
             for byte in key {
-                wrapped = (wrapped * Wrapping(31)) + Wrapping(*byte as i64);
+                wrapped = (wrapped * Wrapping(31)) + Wrapping((*byte as i8) as i64);
```
wprzytula, line 234: Good catch! Casts are always somewhat subtle matter.

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
```diff
@@ -59,6 +60,22 @@ pub struct QueryDefinition {
...
+    pub async fn to_prepared_statement(&self, session: &Arc<Session>) -> Result<PreparedStatement> {
+        let mut statement = session
+            .prepare(&*self.cql)
+            .await
+            .with_context(|| format!("Failed to prepare statement: {}", self.cql))?;
+
+        if let Some(consistency) = self.consistency {
+            statement.set_consistency(consistency);
+        }
+        statement.set_serial_consistency(self.serial_consistency);
```
wprzytula, line 73: I'm not sure if this is what we want about `serial_consistency`. If `QueryDefinition` contains `None::<Consistency>`, then I assume that the intended semantics is that we just want the default consistency. I assume this is the semantics of `None::<SerialConsistency>` as well. The way you did it, however, makes `None::<SerialConsistency>` override the default `Some(SerialConsistency::LocalSerial)` with `None`.

tools/cs_profile_template.yaml
```diff
@@ -0,0 +1,12 @@
...
+table_definition:
+
+keyspace:
+
+keyspace_definition:
+
+queries:
+  test_query:
+    cql:
+    fields: samerow
```
wprzytula: `samerow` or `somerow` ?

wprzytula: I (and the future readers even more) would appreciate comments in the template explaining the semantics and what is expected where.

tools/cs_profile_template.yaml
```diff
@@ -1,12 +0,0 @@
...
-table_definition:
-
-keyspace:
-
-keyspace_definition:
-
-queries:
-  test_query:
-    cql:
-    fields: samerow
```
wprzytula: Hmm, it's gone! So quick! Why was it introduced in the first place then?

tools/util/profiles/cqlstress_blob_profile.yaml
```diff
@@ -0,0 +1,19 @@
+table: test_table
+
+table_definition:
+  CREATE TABLE IF NOT EXISTS test_table (
+    pkey blob PRIMARY KEY,
+    c1 blob,
+    c2 blob,
+    c3 blob
+   );
```
wprzytula, line 9: No `|` here...

tools/util/profiles/cqlstress_blob_profile.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+keyspace: cqlstress_blob_keyspace
+
+keyspace_definition: |
+  CREATE KEYSPACE IF NOT EXISTS cqlstress_blob_keyspace 
+  WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': 1};
```
wprzytula, line 15: ...and I can see `|` here. Why is it so?

tools/util/profiles/cqlstress_blob_profile.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+queries:
+  test_query:
+    cql: insert into cqlstress_blob_keyspace.test_table (pkey, c1, c2, c3) values (?, ?, ?, ?)
```
wprzytula, line 19: Please stick to one case convention in CQL = use CAPITAL LETTERS for CQL keywords.

tools/util/profiles/cs_blob_profile.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+keyspace: cs_blob_keyspace
+
+keyspace_definition: |
+  CREATE KEYSPACE IF NOT EXISTS cs_blob_keyspace 
+  WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': 1};
+
+queries:
+  test_query:
+    cql: insert into cs_blob_keyspace.test_table (pkey, c1, c2, c3) values (?, ?, ?, ?)
```
wprzytula, line 19: ditto

wprzytula (changes requested):

tools/cs_profile_template.yaml
muzarski: I'm sorry. I need to rebase and remove the appearance of this file at all. At first, I wanted to generate test files from this template in CI. This is a remnant that I somehow missed. I think it's better to have yaml files for each type in the repo, so it is clear what CQL statements are executed, what keyspaces/tables we use, etc.

tools/cs_profile_template.yaml
muzarski: `samerow`, see the examples in c-s: https://cassandra.apache.org/doc/stable/cassandra/tools/cassandra_stress.html#profile. IIRC, `queries.<some_query>.fields` parameter is required and accepts either `samerow` or `multirow`. We don't support this parameter in cql-stress yet. From what I remember, this is somehow related to batches (`multirow` would then insert multiple rows in one batch statement). I don't think we should focus on this parameter right now, as it's not commonly used (and AFAIK, the batches are discouraged anyway).

Apart from that, this file is not to be merged anyway. See my other comment where I mention why the file was removed.

tools/util/profiles/cqlstress_blob_profile.yaml
muzarski: It's because of the colons (`:`) contained in the query itself (e.g. `'class':`). This way, they are escaped and not interpreted as another parameter definition. If you remove the pipe symbol, the yaml parser will fail to parse the file. The pipe symbol is not needed for `table_definition`, but I might add it just for the consistency. WDYT?

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
muzarski: Good catch, thanks!

tools/util/profiles/cqlstress_blob_profile.yaml
wprzytula: Yes, please. If it makes sense to always escape CQL statements, then let's do that consistently.

wprzytula (changes requested):
Rebase is still needed to remove the ghostly ephemeral template file. Otherwise, LGTM.

Issue #78, opened 2024-05-09
c-s: Implement column generators for other types


muzarski:
As of now, only `blob` type is supported by the c-s frontend. It's the only type required by existing modes.

With introduction of user profiles, the users should be able to define custom tables with some other types. We need to implement value generators for these types.

roydahan:
While working on that, let's keep in mind that unlike c-s user profile, we want cql-stress to be able to do a data verfication on these types of values.

muzarski:
> While working on that, let's keep in mind that unlike c-s user profile, we want cql-stress to be able to do a data verfication on these types of values.

Yes, but I don't think it's really related to the per-type generation logic itself. Validation heavily depends on `-pop` option, and the distribution that is chosen to sample the partition key seeds from.

I already implemented generators for most of the types, and have it locally (it's just waiting for CLI + runtime to be merged). I made some local tests for these types with `SEQ` distribution (which samples the rows in a deterministic way). The values we generate are the same as in original c-s.

There is one other thing to keep in mind for the future: we want to support UDTs/collections, and so we should implement generators in a way so they can be somehow nested. But I think this can be adjusted in the future, when introducing more complex types to the tool.

PR #83, opened 2024-05-27
c-s: Generators for simple types

muzarski:
Fix: https://github.com/scylladb/cql-stress/issues/78

This PR introduces generator for simple types. These types do not depend on nesting the generators. For more complex types, the `Generator` struct and `ValueGenerator` trait will need to be refactored a bit.

List of the CQL types that I implemented the generators for:
- `text`
- `tinyint`
- `smallint`
- `int`
- `bigint`
- `boolean`
- `float`
- `double`
- `inet`
- `varint`
- `decimal`
- `uuid`

For each type, there is a corresponding test case added to CI.

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/java_generate/values/boolean.rs
+++ b/src/bin/cql-stress-cassandra-stress/java_generate/values/boolean.rs
@@ -1,0 +1,30 @@
+use scylla::frame::response::result::CqlValue;
+
+use crate::java_generate::distribution::Distribution;
+
+use super::{ValueGenerator, ValueGeneratorFactory};
+
+#[derive(Default)]
+pub struct Boolean;
+
+impl ValueGenerator for Boolean {
+    fn generate(
+        &mut self,
+        identity_distribution: &mut dyn Distribution,
+        _size_distribution: &mut dyn Distribution,
+    ) -> CqlValue {
+        // For some reason original c-s returns:
+        // identity_distribution.next_i64() % 1 == 0 which is always true.
+        //
+        // We decided not to follow c-s here and to return truly random boolean.
+        CqlValue::Boolean(identity_distribution.next_i64() % 2 == 1)
+    }
+}
+
+pub struct BooleanFactory;
+
+impl ValueGeneratorFactory for BooleanFactory {
+    fn create(&self) -> Box<dyn ValueGenerator> {
+        Box::<Boolean>::default()
+    }
+}
--- a/src/bin/cql-stress-cassandra-stress/java_generate/values/decimal.rs
+++ b/src/bin/cql-stress-cassandra-stress/java_generate/values/decimal.rs
@@ -1,0 +1,122 @@
+use scylla::frame::{response::result::CqlValue, value::CqlDecimal};
+
+use crate::java_generate::distribution::Distribution;
+
+use super::{ValueGenerator, ValueGeneratorFactory};
+
+#[derive(Default)]
+pub struct Decimal;
+
+impl ValueGenerator for Decimal {
+    fn generate(
+        &mut self,
+        identity_distribution: &mut dyn Distribution,
+        _size_distribution: &mut dyn Distribution,
+    ) -> CqlValue {
+        // The comment of Java's `BigDecimal::valueOf(long)` mentions:
+        // ```Translates a long value into a BigDecimal with a scale of zero.```
+        //
+        // The native representation of `decimal` consists of a `varint` value
+        // and a 32-bit scale/exponent.
+        // This means that we simply need to convert generated i64 value to
+        // `varint`'s native representation and provide a 0-scale.
+        CqlValue::Decimal(CqlDecimal::from_signed_be_bytes_slice_and_exponent(
+            &identity_distribution.next_i64().to_be_bytes(),
+            0,
+        ))
+    }
+}
+
+pub struct DecimalFactory;
+
+impl ValueGeneratorFactory for DecimalFactory {
+    fn create(&self) -> Box<dyn ValueGenerator> {
+        Box::<Decimal>::default()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use bigdecimal::BigDecimal;
+
+    use crate::java_generate::{
+        distribution::fixed::FixedDistribution,
+        values::{decimal::Decimal, Generator, GeneratorConfig},
+    };
+
+    fn bigdecimals_from_i64(values: impl IntoIterator<Item = i64>) -> Vec<BigDecimal> {
+        values.into_iter().map(BigDecimal::from).collect()
+    }
+
+    #[test]
+    fn small_decimal_generator_test() {
+        let config = GeneratorConfig::new(
+            "randomstrC0",
+            None,
+            Some(Box::new(FixedDistribution::new(5))),
+        );
+        let inet_gen = Box::<Decimal>::default();
+        let mut gen = Generator::new(inet_gen, config, String::from("C0"));
+
+        // Values which we test against are generated from c-s.
+        gen.set_seed(0);
+        let results = (0..5)
+            .map(|_| -> BigDecimal { gen.generate().into_cql_decimal().unwrap().into() })
+            .collect::<Vec<_>>();
+        assert_eq!(
+            bigdecimals_from_i64([
+                40527743656,
+                72758341290,
+                51163282362,
+                73862230802,
+                26689604229,
+            ]),
+            results
+        );
+
+        gen.set_seed(0xdeadcafe);
+        let results = (0..5)
+            .map(|_| -> BigDecimal { gen.generate().into_cql_decimal().unwrap().into() })
+            .collect::<Vec<_>>();
+        assert_eq!(
+            bigdecimals_from_i64([
+                26622490754,
+                1431881157,
+                26582476501,
+                62694973673,
+                82585085279,
+            ]),
+            results
+        );
+
+        gen.set_seed(i64::MIN);
+        let results = (0..5)
+            .map(|_| -> BigDecimal { gen.generate().into_cql_decimal().unwrap().into() })
+            .collect::<Vec<_>>();
+        assert_eq!(
+            bigdecimals_from_i64([
+                40527743656,
+                72758341290,
+                51163282362,
+                73862230802,
+                26689604229,
+            ]),
+            results
+        );
+
+        gen.set_seed(i64::MAX);
+        let results = (0..5)
+            .map(|_| -> BigDecimal { gen.generate().into_cql_decimal().unwrap().into() })
+            .collect::<Vec<_>>();
+        assert_eq!(
+            bigdecimals_from_i64([
+                59463298171,
+                52522298470,
+                78786908585,
+                22825301439,
+                15681513599,
+            ]),
+            results
+        );
+    }
+}
--- a/src/bin/cql-stress-cassandra-stress/java_generate/values/inet.rs
+++ b/src/bin/cql-stress-cassandra-stress/java_generate/values/inet.rs
@@ -1,0 +1,122 @@
+use std::net::{IpAddr, Ipv4Addr};
+
+use scylla::frame::response::result::CqlValue;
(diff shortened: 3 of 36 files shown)


tools/util/profiles/cqlstress_bigint_profile.yaml
```diff
@@ -0,0 +1,19 @@
...
+    c1 bigint,
+    c2 bigint,
+    c3 bigint
+   );
+
+keyspace: cqlstress_bigint_keyspace
+
+keyspace_definition: |
+  CREATE KEYSPACE IF NOT EXISTS cqlstress_bigint_keyspace 
+  WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': 1};
```
wprzytula, line 15: Should we be still using `SimpleStrategy` in our internal code, such as in cql-stress implementation?
@piodul @Lorak-mmk

src/bin/cql-stress-cassandra-stress/java_generate/values/boolean.rs
```diff
@@ -0,0 +1,28 @@
...
+
+impl ValueGenerator for Boolean {
+    fn generate(
+        &mut self,
+        _identity_distribution: &mut dyn Distribution,
+        _size_distribution: &mut dyn Distribution,
+    ) -> CqlValue {
+        // For some reason original c-s retunrs:
+        // identity_distribution.next_i64() % 1 == 0 which is always true.
+        CqlValue::Boolean(true)
```
wprzytula, line 18: returns

wprzytula: Wow, no one fixed it in c-s code for 10 years...

wprzytula: @Lorak-mmk WDYT?

src/bin/cql-stress-cassandra-stress/java_generate/values/inet.rs
```diff
@@ -0,0 +1,29 @@
...
+impl ValueGenerator for Inet {
+    fn generate(
+        &mut self,
+        identity_distribution: &mut dyn Distribution,
+        _size_distribution: &mut dyn Distribution,
+    ) -> CqlValue {
+        let v = (identity_distribution.next_i64() as i32).to_be_bytes();
+        CqlValue::Inet(IpAddr::V4(Ipv4Addr::new(v[0], v[1], v[2], v[3])))
+    }
+}
```
wprzytula, line 21: A unit test is missing here, and I claim that this code yields different results than the Java one:
```java
int val = (int) identityDistribution.next();
buf[0] = (byte)(val >>> 24);
buf[1] = (byte)(val >>> 16);
buf[2] = (byte)(val >>> 8);
buf[3] = (byte)val;
return InetAddress.getByAddress(buf);
```
`buf` in Java stores val in LittleEndian, whereas `v` in Rust stores it in BigEndian.

wprzytula: Nit: you can use this impl: 
impl [From](https://doc.rust-lang.org/std/convert/trait.From.html)<[[u8](https://doc.rust-lang.org/std/primitive.u8.html); [4](https://doc.rust-lang.org/std/primitive.array.html)]> for [Ipv4Addr](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html)

wprzytula: If I'm not mistaken and there is a difference in results, how have you tested your impls?

src/bin/cql-stress-cassandra-stress/java_generate/values/varint.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn generate(
+        &mut self,
+        identity_distribution: &mut dyn Distribution,
+        _size_distribution: &mut dyn Distribution,
+    ) -> CqlValue {
+        // BigInt::from_i64 returns Some(). This unwrap is safe.
+        CqlValue::Varint(CqlVarint::from_signed_bytes_be(
+            identity_distribution.next_i64().to_be_bytes().to_vec(),
+        ))
+    }
```
wprzytula, line 20: Comment refers to `unwrap`, but I can't see any.

wprzytula: Nit: `from_signed_bytes_be_slice` is the constructor intended for such a use case - no need for `to_vec` then.

wprzytula, line 20: This is allegedly equivalent code in Java:
```java
    @Override
    public BigInteger generate()
    {
        return BigInteger.valueOf(identityDistribution.next());
    }
```
Can you prove that it's indeed equivalent? Tests are also needed.

src/bin/cql-stress-cassandra-stress/java_generate/values/decimal.rs
```diff
@@ -0,0 +1,33 @@
...
+    ) -> CqlValue {
+        CqlValue::Decimal(
+            // unwrap is safe since BigDecimal::from_i64 returns Some().
+            bigdecimal::BigDecimal::from_i64(identity_distribution.next_i64())
+                .unwrap()
+                .try_into()
+                .expect("Failed to convert i64 to CqlDecimal."),
+        )
+    }
+}
```
wprzytula, line 25: Do we really have to pull in `bigdecimal` dependency for that?
`CqlDecimal` has a constructor: `from_signed_be_bytes_slice_and_exponent` - perhaps it could suffice?

wprzytula: Tests are missing, again.

wprzytula (changes requested):
Unit tests are missing from most commits.

tools/cassandra_stress_ci.py
```diff
@@ -52,3 +52,9 @@ def test_user_blob_type(default_runtime_args, scylla_docker_node,
                         cassandra_stress, cql_stress):
     run_user(runtime_args=default_runtime_args, type_name="blob",
              node=scylla_docker_node, cs=cassandra_stress, cql_stress=cql_stress)
+
+
+def test_user_text_type(default_runtime_args, scylla_docker_node,
+                        cassandra_stress, cql_stress):
+    run_user(runtime_args=default_runtime_args, type_name="text",
+             node=scylla_docker_node, cs=cassandra_stress, cql_stress=cql_stress)
```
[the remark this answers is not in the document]
muzarski: It runs both `cql-stress` and `cassandra-stress`, executes `user` command with corresponding profiles. Based on these profiles, both tools insert some data (by default it's 100 rows - this value is used in CI). Then, we compare that both tables are equal. In other words, we check whether cql-stress generates the same data as original c-s does.

This is why I skipped the unit tests for the types (I saw your comments regarding that). However, I can still add the unit tests for the types if necessary. WDYT?

muzarski:
v2: rebased on master

src/bin/cql-stress-cassandra-stress/java_generate/values/boolean.rs
Lorak-mmk: Is it that important to be bug-compatible with cql-stress? Imo we should do it the right way, I doubt anyone depends on this bug.

tools/util/profiles/cqlstress_bigint_profile.yaml
Lorak-mmk: I think it should not be the default (and it is used in all profiles right now, so that should be changed) but there should be some test cases for it as long as it exists in Scylla / C*

muzarski: Notice that these profiles are only used for tests.

BTW, related question: as of now, the cluster used in the CI consists of only one node. Maybe it's something to be changed? Maybe we should use a 3-node cluster. Then it would make sense to use/test some other replication strategies.

src/bin/cql-stress-cassandra-stress/java_generate/values/inet.rs
muzarski: You can run the test from CI:
```bash
# Setup a cluster
docker compose -f docker/scylla_test/compose.yml up -d --wait

# Run the test
pytest -s ./tools/cassandra_stress_ci.py -k test_user_inet_type

# You can examine the tables:
cqlsh
> SELECT * FROM cs_inet_keyspace.test_table ;
> SELECT * FROM cqlstress_inet_keyspace.test_table ;
```

tools/util/profiles/cqlstress_bigint_profile.yaml
Lorak-mmk: Even if it's  only tests we should use NTS by default.
Yes, bigger cluster in CI sounds good

src/bin/cql-stress-cassandra-stress/java_generate/values/decimal.rs
muzarski: > Do we really have to pull in `bigdecimal` dependency for that? `CqlDecimal` has a constructor: `from_signed_be_bytes_slice_and_exponent` - perhaps it could suffice?

I removed the dependency and used the constructor you mentioned. Added a comment that justifies values provided to the constructor.

src/bin/cql-stress-cassandra-stress/java_generate/values/varint.rs
muzarski: > Comment refers to `unwrap`, but I can't see any.

It's a remnant that I forgot to remove. First version would use a `num_bigint` crate and its `BigInt` struct. However, the dependency is not needed since we can easily represent `i64` in a native `varint` representation (a signed integer's representation in big-endian order bytes).

src/bin/cql-stress-cassandra-stress/java_generate/values/boolean.rs
muzarski: I changed it so we generate a truly random `boolean` value. Removed the compat test cases for booleans from CI, since we are not compatible with c-s in this case.

muzarski:
v2: I addressed some of the review comments in the code.

There are still some things to be discussed. ping @wprzytula

tools/cassandra_stress_ci.py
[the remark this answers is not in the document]
wprzytula: If compatibility with c-s is our only source of truth about correctness, then I think that these tests are enough. OTOH, if a writing a particular unit test is a low-hanging fruit, then it's better to have one.

src/bin/cql-stress-cassandra-stress/java_generate/values/inet.rs
wprzytula: You're right, reading it now I have no idea why I reasoned it's LE in Java.

tools/cassandra_stress_ci.py
```diff
@@ -52,3 +52,69 @@ def test_user_blob_type(default_runtime_args, scylla_docker_node,
...
+
+
+def test_user_tinyint_type(default_runtime_args, scylla_docker_node,
+                           cassandra_stress, cql_stress):
+    run_user(runtime_args=default_runtime_args, type_name="tinyint",
+             node=scylla_docker_node, cs=cassandra_stress, cql_stress=cql_stress)
+
```
wprzytula, line 67: Please mention here as well why test for boolean is missing (purposeful incompatibility introduced).

wprzytula (approved):

tools/util/profiles/cqlstress_bigint_profile.yaml
muzarski: Created an issue: https://github.com/scylladb/cql-stress/issues/88

muzarski:
v3: implemented unit tests for the types.

wprzytula (approved):

muzarski:
Rebased on main

wprzytula (approved):

Issue #85, opened 2024-06-10
c-s: Tool creates `counter1` table when regular workloads are being run


muzarski:
Ref: https://github.com/scylladb/scylladb/issues/19107#issuecomment-2155038858

Cassandra-stress frontend always creates `standard1` and `counter1` tables for all workloads (`user` is exception). At the time of implementing this, I was sure that original c-s does the same - apparently I was wrong.

In general, the tables we should create for each workload:
- `write`/`read` - `standard1` table
- `counter`/`counter_read` - `counter1` table
- `mixed` - both `standard1` and `counter1` tables. Better idea might be to create the tables that will be used by the commands provided by the user - e.g. when user provides `ratio(write=1,read=2)`, we should not create `counter1` table since it won't be used
- `user` - a table defined in yaml file with `table` and `table_definition` parameters

PR #86, opened 2024-06-10
c-s: create correct tables based on workload

muzarski:
fix: https://github.com/scylladb/cql-stress/issues/85

List of tables created for each command:
- read/write -> `keyspace1.standard1` table
- counter_read/counter_write -> `keyspace1.counter1` table
- mixed -> both `keyspace1.[standard1/counter1]` tables
- user -> keyspace and table provided in user profile yaml file

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/main.rs
+++ b/src/bin/cql-stress-cassandra-stress/main.rs
@@ -138,7 +138,10 @@ async fn prepare_run(
     let session = builder.build().await?;
     let session = Arc::new(session);
 
-    create_schema(&session, &settings).await?;
+    settings
+        .create_schema(&session)
+        .await
+        .context("Failed to create schema")?;
 
     let duration = settings.command_params.common.duration;
 
@@ -161,50 +164,6 @@ async fn prepare_run(
         // TODO: adjust when -errors option is supported
         max_retries_per_op: 9,
     })
-}
-
-async fn create_schema(session: &Session, settings: &CassandraStressSettings) -> Result<()> {
-    match settings.command {
-        #[cfg(feature = "user-profile")]
-        Command::User => {
-            // 'user' command provided. This unwrap is safe.
-            settings
-                .command_params
-                .user
-                .as_ref()
-                .unwrap()
-                .create_schema(session)
-                .await?;
-        }
-        _ => {
-            session
-                .query(settings.schema.construct_keyspace_creation_query(), ())
-                .await?;
-            session
-                .use_keyspace(&settings.schema.keyspace, true)
-                .await?;
-            session
-                .query(
-                    settings
-                        .schema
-                        .construct_table_creation_query(&settings.column.columns),
-                    (),
-                )
-                .await
-                .context("Failed to create standard table")?;
-            session
-                .query(
-                    settings
-                        .schema
-                        .construct_counter_table_creation_query(&settings.column.columns),
-                    (),
-                )
-                .await
-                .context("Failed to create counter table")?;
-        }
-    }
-
-    Ok(())
 }
 
 async fn create_operation_factory(
--- a/src/bin/cql-stress-cassandra-stress/settings/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/mod.rs
@@ -60,26 +60,37 @@ impl CassandraStressSettings {
             return user.create_schema(session).await;
         }
 
-        session
-            .query(self.schema.construct_keyspace_creation_query(), ())
-            .await?;
+        if matches!(self.command, Command::Write | Command::CounterWrite) {
+            session
+                .query(self.schema.construct_keyspace_creation_query(), ())
+                .await?;
+        }
+
         session.use_keyspace(&self.schema.keyspace, true).await?;
-        session
-            .query(
-                self.schema
-                    .construct_table_creation_query(&self.column.columns),
-                (),
-            )
-            .await
-            .context("Failed to create standard table")?;
-        session
-            .query(
-                self.schema
-                    .construct_counter_table_creation_query(&self.column.columns),
-                (),
-            )
-            .await
-            .context("Failed to create counter table")?;
+
+        match self.command {
+            Command::Write => {
+                session
+                    .query(
+                        self.schema
+                            .construct_table_creation_query(&self.column.columns),
+                        (),
+                    )
+                    .await
+                    .context("Failed to create standard table")?;
+            }
+            Command::CounterWrite => {
+                session
+                    .query(
+                        self.schema
+                            .construct_counter_table_creation_query(&self.column.columns),
+                        (),
+                    )
+                    .await
+                    .context("Failed to create counter table")?;
+            }
+            _ => (),
+        }
 
         Ok(())
     }
--- a/src/bin/cql-stress-cassandra-stress/settings/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/mod.rs
@@ -4,6 +4,7 @@ use std::iter::Iterator;
 mod command;
 mod option;
 mod param;
+use anyhow::Context;
 use anyhow::Result;
 
 #[cfg(test)]
@@ -17,6 +18,7 @@ pub use command::OpWeight;
 pub use command::OperationRatio;
 pub use option::ThreadsInfo;
 use regex::Regex;
+use scylla::Session;
 
 use crate::settings::command::print_help;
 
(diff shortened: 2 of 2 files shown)


wprzytula (changes requested):
For substantially better review experience (clear diff between old and new logic), please extract another commit that makes the `create_schema` function a method (without any changes to its logic) (pure refactor commit), and then proceed with the commit that changes logic only.

src/bin/cql-stress-cassandra-stress/settings/command/mod.rs
```diff
@@ -142,17 +142,31 @@ impl CommandParams {
...
-        session
-            .query(
-                schema.construct_counter_table_creation_query(&column.columns),
-                (),
-            )
-            .await
-            .context("Failed to create counter table")?;
 
         Ok(())
     }
```
wprzytula, line 172: I believe the logic would be much more readable if you separated table conditions from table creation, e.g.:
```rust
let (create_standard_table, create_counter_table) = match (self.counter, self.mixed) {
  // conditions
}

// Actual table creation code
if create_standard_table {
...
}

if create_counter_table {
...
}
```

muzarski: I love this! I'll apply the suggestion.

muzarski:
v2: addressed review comment

wprzytula (approved):

roydahan:
> * mixed -> both `keyspace1.[standard1/counter1]` tables

@muzarski the defined behavior is incorrect.
Actually mixed should be the same as "read/write" and run on table keyspace1.standard1.
The difference is that mixed should do 50/50 reads/writes ops.

src/bin/cql-stress-cassandra-stress/settings/command/mod.rs
```diff
@@ -123,6 +126,48 @@ impl CommandParams {
...
+        #[cfg(feature = "user-profile")]
+        if let Some(user) = &self.user {
+            return user.create_schema(session).await;
+        }
+
+        let (create_standard_table, create_counter_table) = match (&self.mixed, &self.counter) {
+            (None, None) => (true, false),
+            (None, Some(_)) => (false, true),
+            (Some(_), _) => (true, true),
+        };
```
roydahan, line 145: This one is incorrect definition

roydahan (changes requested):
Need to change mixed behavior

muzarski:
> > * mixed -> both `keyspace1.[standard1/counter1]` tables
> 
> @muzarski the defined behavior is incorrect. Actually mixed should be the same as "read/write" and run on table keyspace1.standard1. The difference is that mixed should do 50/50 reads/writes ops.

`cassandra-stress help mixed` says:
```
ratio([read=?][write=?][counter_write=?][counter_read=?]): Specify the ratios for operations to perform; e.g. (read=2,write=1) will perform 2 reads for each write
      read=? (default=1)                       Performs this many READ operations out of total
      write=? (default=1)                      Performs this many WRITE operations out of total
      counter_write=?                          Performs this many COUNTER_WRITE operations out of total
      counter_read=?                           Performs this many COUNTER_READ operations out of total
```

This means that `counter_write` and `counter_read` operations can be performed as well. Thus, the table `keyspace1.counter1` is being used as well.

muzarski:
> The difference is that mixed should do 50/50 reads/writes ops.

The 50/50 reads/writes is the default behaviour, but can be configured via `ratio` parameter (so the tool performs operations on counter tables as well).

roydahan:
Ok, I understand now.
So, as long as the default ratio is only "read,write" it should be fine.

roydahan (dismissed):

roydahan:
@soyacz I asked @muzarski to merge and release so we can retest the tablets one without even creating the table.
The "mixed" could support counters workload but you must explictily state a ratio that include counter_write, counter_read.

roydahan:
@muzarski rethinking of it, AFAIR in c-s we don't create the tables in "mixed" command, it assumes that one already ran a "write" command before that in order to prepare for it.
So, unless it's an improvement we're adding on top of c-s, cql-stress shouldn't do it as well.

muzarski:
> order to prepare for it

Oh, that's interesting. But it would make sense, since `mixed` allows to execute `read` operation (which requires a populated DB). Let me verify this, and I'll adjust the logic properly for cql-stress.

muzarski:
v3:
List of tables created for each command:
- `write` -> `keyspace1.standard1`
- `counter_write` -> `keyspace1.counter1`
- `user` -> based on profile
- `read/counter_read/mixed` -> no table created

@roydahan Please confirm that it makes sense now.

roydahan (approved):

fruch:
@muzarski 

now we don't create the schema...

but the code still tries to prepare statements on the non-existing keyspace/table..., which is even worse, mixed can't be use now...

```
TAG: loader_idx:4-cpu_idx:0-keyspace_idx:1
******************** Stress Settings ********************
Command:
  Type: mixed
  Count: -1
  Duration: 48000 SECONDS
  No Warmup: true
  Consistency Level: Quorum
  Serial Consistency Level: Serial
  Truncate: NEVER
  Target Uncertainty: not applicable
  Key Size (bytes): 10
  Counter Increment Distibution: FIXED(1)
Command ratios: {read=1,write=1}
Command clustering distribution: GAUSSIAN(1..10,mean=5.5,stdev=1.5)
Rate:
  Thread count: 300
  OpsPer Sec: 8750
  Coordinated-Omission-Fixed latencies: true
Mode:
  Compression: None
  Pool size: PerShard(1)
Node:
  Nodes: ["10.4.1.151", "10.4.3.241", "10.4.3.226"]
  Is White List: false
  Datacenter: None
Schema:
  Keyspace: keyspace1
  Replication Strategy Options: {"replication_factor": "3", "class": "NetworkTopologyStrategy"}
  Table Compression: None
  Table Compaction Options: {}
Column:
  Column names: ["C0", "C1", "C2", "C3", "C4", "C5", "C6", "C7"]
  Size distribution: FIXED(128)
Population:
  Partition key seed distribution: GAUSSIAN(1..650000000,mean=325000000,stdev=6500000)

Error: Failed to prepare benchmark

Caused by:
    0: Failed to prepare statement
    1: Database returned an error: The query is syntactically correct but invalid, Error message: unconfigured table counter1
```

muzarski:
> @muzarski rethinking of it, AFAIR in c-s we don't create the tables in "mixed" command, it assumes that one already ran a "write" command before that in order to prepare for it.
> 
> So, unless it's an improvement we're adding on top of c-s, cql-stress shouldn't do it as well.

@fruch WDYT?

muzarski:
I can reverse the changes regarding mixed workload so it creates the schema.

muzarski:
Looking at the code of original c-s:
```java
    public void maybeCreateKeyspaces()
    {
        if (command.type == Command.WRITE || command.type == Command.COUNTER_WRITE)
            schema.createKeySpaces(this);
        else if (command.type == Command.USER)
            ((SettingsCommandUser) command).profile.maybeCreateSchema(this);
    }
```

Keyspace and tables are not created for mixed workloads. @fruch Please try to run the `write` workload firstly to create a schema. Or if this really annoying and you think that we should create schema during `mixed` command as well, please open an issue about it so we can discuss it there.

fruch:
> Looking at the code of original c-s:
> ```java
>     public void maybeCreateKeyspaces()
>     {
>         if (command.type == Command.WRITE || command.type == Command.COUNTER_WRITE)
>             schema.createKeySpaces(this);
>         else if (command.type == Command.USER)
>             ((SettingsCommandUser) command).profile.maybeCreateSchema(this);
>     }
> ```
> 
> Keyspace and tables are not created for mixed workloads. @fruch Please try to run the `write` workload firstly to create a schema. Or if this really annoying and you think that we should create schema during `mixed` command as well, please open an issue about it so we can discuss it there.

We did run the write workload before, but the mixed workload shouldn't do any operations with counters, if we want counter there is a command for that.

Now tablets doesn't support counters, and we need that that read/write/mixed won't even create the counter schema, and not do any counter operations

muzarski:
> > Looking at the code of original c-s:
> > ```java
> >     public void maybeCreateKeyspaces()
> >     {
> >         if (command.type == Command.WRITE || command.type == Command.COUNTER_WRITE)
> >             schema.createKeySpaces(this);
> >         else if (command.type == Command.USER)
> >             ((SettingsCommandUser) command).profile.maybeCreateSchema(this);
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Keyspace and tables are not created for mixed workloads. @fruch Please try to run the `write` workload firstly to create a schema. Or if this really annoying and you think that we should create schema during `mixed` command as well, please open an issue about it so we can discuss it there.
> 
> We did run the write workload before, but the mixed workload shouldn't do any operations with counters, if we want counter there is a command for that.
> 
> Now tablets doesn't support counters, and we need that that read/write/mixed won't even create the counter schema, and not do any counter operations

Oh, I see. We try to prepare a statement operating on `counter1` table. This is indeed a bug. I'll try to fix that ASAP.

PR #87, opened 2024-06-12
c-s: Implement query/operation sampler

muzarski:
Fix: https://github.com/scylladb/cql-stress/issues/80

This PR extends a CLI for user command by `ops()` and `clustering=` parameters.

The semantics of parameters:
- `ops` -> defines a ratio with which the operation/queries will be
    sampled. For example `ops(foo=1,bar=2)` means that the tool will
    approximately execute 2 `bar` queries per 1 `foo` query. These queries
    must be defined in the profile yaml.
- `clustering` -> a distribution. The tool will sample a value from
  this distribution to decide how many times to execute
  current operation/query (which was sampled based on `ops`). The value
  will be treated as a counter. When counter reaches 0, new operation
  is sampled, and so, the new value of counter is sampled as well.

The struct that will make use of these parameters is `OperationSampler`. At first, I wanted to implement it in a way so it can be used for both `user` and `mixed` commands. The reason for this is that `mixed` command supports two analogous parameters: `ratio` and `clustering`. However, due to the Rust language limitations it's really hard (or even impossible) to implement the operation sampler this way. `CassandraStressOperation` trait declares async methods, thus it is not object-safe, and so we cannot create trait objects based on this trait (`Box<dyn CassandraStressOperation>`).

I decided to leave the code for `mixed` command as is for now.

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/operation/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mod.rs
@@ -11,6 +11,8 @@ use cql_stress::configuration::Operation;
 use cql_stress::configuration::OperationContext;
 use cql_stress::configuration::OperationFactory;
 use cql_stress::make_runnable;
+#[cfg(feature = "user-profile")]
+use rand_distr::{Distribution as _, WeightedIndex};
 use scylla::Session;
 use std::future::Future;
 use std::num::Wrapping;
@@ -26,6 +28,8 @@ use scylla::{
 #[cfg(feature = "user-profile")]
 pub use user::UserOperationFactory;
 
+#[cfg(feature = "user-profile")]
+use crate::java_generate::distribution::{Distribution, DistributionFactory};
 use crate::settings::CassandraStressSettings;
 use crate::stats::ShardedStats;
 
@@ -308,5 +312,66 @@ impl RowValidator for ExistsRowValidator {
         // successfully extracting the first row from the query result.
         let _first_row = extract_first_row_from_query_result(&query_result)?;
         Ok(())
+    }
+}
+
+/// A sampler created based on a ratio map and a counter distribution.
+///
+/// How the sampler works?
+/// One iteration consists of:
+/// - sampling an item based on ratio map. `current_item_index` is sampled from `item_index_dist`.
+///   The item can then be retrieved via this index from `items` vector.
+/// - sampling a counter which says how many times to return the current item.
+///   The counter is sampled from `counter_dist` distribution.
+///
+/// The user then can sample the items via `sample` or `previous_sample` method.
+///
+/// The `sample` method will decrease the counter by 1, and return current item.
+/// If the counter reaches 0, new iteration starts.
+///
+/// The `previous_sample` method returns a current item without decreasing the counter.
+/// This is helpful when the user wants to, for example, retry an operation that was
+/// sampled before, but failed for some reason.
+#[cfg(feature = "user-profile")]
+struct OperationSampler<T> {
+    counter_dist: Box<dyn Distribution>,
+    items: Vec<T>,
+    item_index_dist: WeightedIndex<f64>,
+    current_item_remaining: u8,
+    current_item_index: usize,
+}
+
+#[cfg(feature = "user-profile")]
+impl<T> OperationSampler<T> {
+    pub fn new(
+        weights: impl Iterator<Item = (T, f64)>,
+        counter_dist_factory: &dyn DistributionFactory,
+    ) -> Self {
+        let (items, weights): (Vec<_>, Vec<_>) = weights.unzip();
+        // We verify the ratio properties during parsing.
+        let item_index_dist = WeightedIndex::new(weights).unwrap_or_else(|err| {
+            panic!("Failed to create a WeightedIntex from provided ratios: {err}")
+        });
+
+        Self {
+            counter_dist: counter_dist_factory.create(),
+            items,
+            item_index_dist,
+            current_item_remaining: 0,
+            current_item_index: 0,
+        }
+    }
+
+    pub fn sample(&mut self) -> &T {
+        if self.current_item_remaining == 0 {
+            self.current_item_index = self.item_index_dist.sample(&mut rand::thread_rng());
+            self.current_item_remaining = (self.counter_dist.next_i64() as u8).max(1);
+        }
+        self.current_item_remaining -= 1;
+        &self.items[self.current_item_index]
+    }
+
+    pub fn previous_sample(&self) -> &T {
+        &self.items[self.current_item_index]
     }
 }
--- a/src/bin/cql-stress-cassandra-stress/operation/user.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/user.rs
@@ -13,15 +13,16 @@ use anyhow::Result;
 
 use crate::{
     java_generate::{
-        distribution::Distribution,
+        distribution::{Distribution, DistributionFactory},
         values::{Generator, GeneratorConfig, ValueGeneratorFactory},
     },
-    settings::CassandraStressSettings,
+    settings::{CassandraStressSettings, OpWeight},
     stats::ShardedStats,
 };
 
 use super::{
     row_generator::RowGenerator, CassandraStressOperation, CassandraStressOperationFactory,
+    OperationSampler,
 };
 
 const SEED_STR: &str = "seed for stress";
@@ -70,49 +71,8 @@ impl CassandraStressOperationFactory for UserDefinedOperationFactory {
     }
 }
 
-/// A struct that samples the operations.
-/// TODO: For now, this is a simple round-robin sampler.
-///       Adjust it, when `ops()` and `clustering=` parameters are supported.
-///       This will need to be somehow unified with the sampler of `mixed` operation.
-///       I think the sampler could cache the row associated with current operation as well.
-struct OperationSampler {
-    op_map: HashMap<String, UserDefinedOperation>,
-    op_keys: Vec<String>,
-    current_operation_index: usize,
-}
-
-impl OperationSampler {
-    fn from_operation_map(op_map: HashMap<String, UserDefinedOperation>) -> Self {
-        let op_keys = op_map.keys().cloned().collect::<Vec<_>>();
-        let current_operation_index = 0;
-
-        Self {
-            op_map,
-            op_keys,
-            current_operation_index,
-        }
-    }
-
-    fn sample(&mut self) -> &UserDefinedOperation {
-        // op_keys is a vector of keys of op_map. This unwrap is safe.
-        let sample = self
-            .op_map
-            .get(&self.op_keys[self.current_operation_index])
-            .unwrap();
-        self.current_operation_index = (self.current_operation_index + 1) % self.op_keys.len();
-        sample
-    }
-
-    fn sample_cached(&self) -> &UserDefinedOperation {
-        // op_keys is a vector of keys of op_map. This unwrap is safe.
-        self.op_map
-            .get(&self.op_keys[self.current_operation_index])
-            .unwrap()
-    }
-}
-
 pub struct UserOperation {
-    sampler: OperationSampler,
+    sampler: OperationSampler<UserDefinedOperation>,
     workload: RowGenerator,
     stats: Arc<ShardedStats>,
     max_operations: Option<u64>,
@@ -130,7 +90,7 @@ impl UserOperation {
         }
 
         let (op, row) = match &mut self.cached_row {
-            Some(cached_row) => (self.sampler.sample_cached(), cached_row),
+            Some(cached_row) => (self.sampler.previous_sample(), cached_row),
             None => {
                 let op = self.sampler.sample();
                 let row = self.cached_row.insert(op.generate_row(&mut self.workload));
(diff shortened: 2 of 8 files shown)


src/bin/cql-stress-cassandra-stress/settings/param/types.rs
```diff
@@ -298,9 +299,86 @@ impl Parsable for ConnectionsPerShard {
...
+    fn do_parse(s: &str) -> Result<HashMap<String, f64>> {
+        // Remove wrapping parenthesis.
+        let arg = {
+            let mut chars = s.chars();
+            anyhow::ensure!(
+                chars.next() == Some('(') && chars.next_back() == Some(')'),
+                "Invalid ratio specification: {}",
+                s
+            );
+            chars.as_str()
```
wprzytula, line 340: Wow, what a nice API of `str::Chars`!

src/bin/cql-stress-cassandra-stress/settings/param/types.rs
```diff
@@ -298,9 +299,86 @@ impl Parsable for ConnectionsPerShard {
...
+        };
+
+        // A set to ensure that items are unique.
+        let mut item_set = HashSet::<&str>::new();
+        // Verify that sum of weights is non-zero.
+        let mut sum = 0f64;
+        let weights_map = arg
+            .split(',')
+            .map(|s| -> Result<(String, f64)> {
```
wprzytula, line 349: TIL: it's possible to explicitly state return type of a closure. Thanks.

src/bin/cql-stress-cassandra-stress/settings/param/types.rs
```diff
@@ -298,9 +299,86 @@ impl Parsable for ConnectionsPerShard {
...
+
+    fn do_parse(s: &str) -> Result<HashMap<String, f64>> {
+        // Remove wrapping parenthesis.
+        let arg = {
+            let mut chars = s.chars();
+            anyhow::ensure!(
+                chars.next() == Some('(') && chars.next_back() == Some(')'),
+                "Invalid ratio specification: {}",
+                s
+            );
```
wprzytula, line 339: Note that if this `ensure!` fails, the resulting error will contain a message with double `Invalid ratio specification: Invalid ratio specification: ...`, because the outer `do_parse` prepends such header too.

src/bin/cql-stress-cassandra-stress/settings/param/types.rs
```diff
@@ -398,4 +476,26 @@ mod tests {
             assert!(DistributionTestType::parse(input).is_err());
         }
     }
+
+    #[test]
+    fn ratio_map_param_test() {
+        let good_test_cases = ["(foo=1)", "(foo=1.2,bar=21,baz=0.5)"];
```
wprzytula, line 482: Please add an item with `0` ratio to showcase that it's a correct input.

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
```diff
@@ -54,6 +54,7 @@ impl QueryDefinitionYaml {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
 pub struct QueryDefinition {
     pub cql: String,
     pub consistency: Option<Consistency>,
```
wprzytula, line 62: If deriving `PartialEq`, `Eq` should be derived as well unless there is a reason not to.

src/bin/cql-stress-cassandra-stress/settings/command/test_user_profile_yamls/full_profile.yaml
```diff
@@ -16,7 +16,7 @@ table_definition:
 queries:
   ins:
     cql: insert into standard1 (pkey, ckey, c1) values (?, ?, ?)
-    consistencyLevel: local
+    consistencyLevel: local_one
     serialConsistencyLevel: local_serial
```
wprzytula, line 20: Oops, how did it not cause some tests to fail?

src/bin/cql-stress-cassandra-stress/operation/user.rs
```diff
@@ -198,7 +198,7 @@ impl UserOperationFactory {
         );
 
         let mut statements_map = HashMap::new();
-        for (q_name, q_def) in query_definitions {
+        for (q_name, (q_def, _weigth)) in query_definitions {
```
wprzytula: `weigth` -> `weight`

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
```diff
@@ -100,7 +102,10 @@ pub struct UserParams {
     pub keyspace_definition: Option<String>,
     pub table: String,
     pub table_definition: Option<String>,
-    pub queries: HashMap<String, QueryDefinition>,
+    // Maps a query name to query definition and a ratio with which
+    // this query will be sampled.
+    pub queries_payload: HashMap<String, (QueryDefinition, f64)>,
```
wprzytula, line 107: WDYT about introducing a type alias `type OpWeight = f64` ? Would it impact readability positively or negatively?

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
```diff
@@ -147,32 +152,43 @@ impl UserParams {
...
+                        .remove(&query_name)
+                        .ok_or_else(|| {
+                            anyhow::anyhow!("Unrecognized query name in ratio map: {}", query_name)
+                        })?
+                        .into_query_definition()
+                        .context("Failed to parse query definition")?;
+
+                    Ok((query_name, (query_def, weight)))
+                },
+            )
```
wprzytula, line 174: Is it an intended semantics change that now queries present in `queries` but without weights specified in `ratio` are ignored (i.e. not put in the returned map)?

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -310,3 +314,63 @@ impl RowValidator for ExistsRowValidator {
...
+#[cfg(feature = "user-profile")]
+impl<T> OperationSampler<T> {
+    pub fn new(
+        weights: impl Iterator<Item = (T, f64)>,
+        counter_dist_factory: &dyn DistributionFactory,
+    ) -> Self {
+        let (items, weights): (Vec<_>, Vec<_>) = weights.unzip();
+        // We verify the ratio properties during parsing.
+        let item_index_dist = WeightedIndex::new(weights)
+            .expect("Failed to create a WeightedIntex from provided ratios");
```
wprzytula, line 353: Better not to hide the specific error.

wprzytula:
```suggestion
        let item_index_dist = WeightedIndex::new(weights)
            .unwrap_or_else(|err| panic!("Failed to create a WeightedIntex from provided ratios: {}", err);
```

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -310,3 +314,63 @@ impl RowValidator for ExistsRowValidator {
...
+            self.current_item_index = self.item_index_dist.sample(&mut rand::thread_rng());
+            self.current_item_remaining = (self.counter_dist.next_i64() as u8).max(1);
+        }
+        self.current_item_remaining -= 1;
+        &self.items[self.current_item_index]
+    }
+
+    pub fn sample_cached(&self) -> &T {
+        &self.items[self.current_item_index]
+    }
```
wprzytula, line 375: I would rather name this function `previous_sample`, because actually no sampling happens here - just returning what was returned previously.
[_Probability sampling refers to the selection of a sample from a population, when this selection is based on the principle of randomization, that is, random selection or chance._](https://www150.statcan.gc.ca/n1/edu/power-pouvoir/ch13/prob/5214899-eng.htm)
...and there is no random choice here.

src/bin/cql-stress-cassandra-stress/operation/user.rs
```diff
@@ -275,12 +237,15 @@ impl UserOperationFactory {
 
 impl OperationFactory for UserOperationFactory {
     fn create(&self) -> Box<dyn Operation> {
+        // We parsed a user command. This unwrap is safe.
+        let user_profile = self.settings.command_params.user.as_ref().unwrap();
+
```
wprzytula, line 242: I really dislike this design that requires us to `unwrap` everywhere.

src/bin/cql-stress-cassandra-stress/operation/user.rs
```diff
@@ -159,10 +119,11 @@ pub struct UserOperationFactory {
     pk_seed_distribution: Arc<dyn Distribution>,
     stats: Arc<ShardedStats>,
     table_metadata: Table,
-    statements_map: HashMap<String, PreparedStatement>,
+    queries_payload: HashMap<String, (PreparedStatement, f64)>,
     pk_generator_factory: Box<dyn ValueGeneratorFactory>,
     column_generator_factories: Vec<Box<dyn ValueGeneratorFactory>>,
     max_operations: Option<u64>,
+    settings: Arc<CassandraStressSettings>,
 }
```
wprzytula, line 127: At least in this commit, it's not clear to me what are these `settings` and why we add them here.

wprzytula (changes requested):

src/bin/cql-stress-cassandra-stress/operation/user.rs
wprzytula: It actually was the last commit that could explain this. So why?

src/bin/cql-stress-cassandra-stress/settings/command/test_user_profile_yamls/full_profile.yaml
muzarski: Previously, we only tested it parsing it to `String`. Now we test parsing into `scylla::Consistency` type as well.

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
muzarski: Yes, such operations won't be sampled anyway so there is no point in holding info about them. This just means that the user defined an operation in the yaml file, but decided not to use it during the runtime. The use case for this might be to have a single yaml file for multiple stress scenarios (e.g. each scenario executes only one operation from the yaml file).

For a moment, I thought that maybe we should fail parsing if there is an operation present in `queries` but not in `queries_ratio`. However, I think this could be annoying for users who want to define a single yaml file and reuse it for multiple stress runs.

muzarski: + This is the behaviour of original c-s.

src/bin/cql-stress-cassandra-stress/operation/user.rs
muzarski: > it's not clear to me what are these `settings` and why we add them here.

I just want to pass the `clustering` object from the settings which is of type `Box<dyn DistributionFactory>`. Unfortunately, we cannot clone is since it is not `Sized`. We also cannot just move the `clustering` from the settings since it's behind a shared reference (`Arc`). Do you have any idea on how to pass the distribution factory here (instead of passing whole `CassandraStressSettings` structure)?

muzarski: Maybe I could just wrap it with `Arc`, instead of `Box`. Then we could have a shared reference to the distribution factory. Let me try this out.

muzarski: > Maybe I could just wrap it with `Arc`, instead of `Box`. Then we could have a shared reference to the distribution factory. Let me try this out.

Yup, this will work. This also gets rid of the unwrap you mentioned in the other comment.

src/bin/cql-stress-cassandra-stress/operation/user.rs
muzarski: I got rid of this by not passing whole `settings` object to the factory. However, we will still discuss the design later offline.

muzarski:
v2: addressed review comments

src/bin/cql-stress-cassandra-stress/operation/user.rs
```diff
@@ -159,10 +119,11 @@ pub struct UserOperationFactory {
     pk_seed_distribution: Arc<dyn Distribution>,
     stats: Arc<ShardedStats>,
     table_metadata: Table,
-    statements_map: HashMap<String, PreparedStatement>,
+    queries_payload: HashMap<String, (PreparedStatement, f64)>,
     pk_generator_factory: Box<dyn ValueGeneratorFactory>,
```
wprzytula, line 123: `f64` -> `OpWeight` ?

src/bin/cql-stress-cassandra-stress/settings/param/types.rs
```diff
@@ -298,9 +299,85 @@ impl Parsable for ConnectionsPerShard {
...
+    fn do_parse(s: &str) -> Result<HashMap<String, f64>> {
+        // Remove wrapping parenthesis.
+        let arg = {
+            let mut chars = s.chars();
+            anyhow::ensure!(
+                chars.next() == Some('(') && chars.next_back() == Some(')'),
+                "List of item weights should be wrapped with parenthesis",
+            );
+            chars.as_str()
+        };
```
wprzytula, line 340: parenthesis -> parentheses

wprzytula (changes requested):

wprzytula (approved):

PR #89, opened 2024-06-18
CI: fix clippy issues

muzarski:
There were two issues clippy was complaining about:
- use of legacy numeric types (i.e. `std::{u32. u64}`)
- unused field in `Timeseries{Read/Write}Config` structs (no idea why this wasn't recognized before)

This PR fixes these issues.

### DIFF
--- a/src/bin/cql-stress-scylla-bench/workload/timeseries_read.rs
+++ b/src/bin/cql-stress-scylla-bench/workload/timeseries_read.rs
@@ -28,7 +28,8 @@ struct TimeseriesRead {
 
 #[derive(Clone)]
 pub struct TimeseriesReadConfig {
-    pub partition_offset: i64,
+    // FIXME: why is this not used??
+    pub _partition_offset: i64,
     pub pks_per_generation: u64,
     pub cks_per_pk: u64,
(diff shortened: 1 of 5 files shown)


src/bin/cql-stress-scylla-bench/workload/timeseries_read.rs
```diff
@@ -28,7 +28,8 @@ struct TimeseriesRead {
 
 #[derive(Clone)]
 pub struct TimeseriesReadConfig {
-    pub partition_offset: i64,
+    // FIXME: why is this not used??
+    pub _partition_offset: i64,
     pub pks_per_generation: u64,
     pub cks_per_pk: u64,
     pub start_nanos: u64,
```
wprzytula, line 35: Perhaps it's worth immediate investigation, not to let it fall into oblivion of numerous FIXMEs?

wprzytula (approved):

src/bin/cql-stress-scylla-bench/workload/timeseries_read.rs
wprzytula: @muzarski please don't forget about this.

Issue #91, opened 2024-06-19
c-s: Implement full row insertion logic for user-profiles


muzarski:
Original c-s allows the user to define custom operations/queries via `user` command. However, it also predefines two operations - one of them being `insert` operation. It simply generates a row based on table metadata, and inserts it. Row values can then be validated with [`validate`](https://github.com/scylladb/cql-stress/issues/90) operation.

Things to do:
- adjust `user` command CLI to support `insert` operation
- implement `insert` operation logic:
  - generate a row
  - insert the row to DB

Notice that `insert/validate` operations are analogous to `write/read` operations for the predefined workloads. The difference is that they allow users to work with custom schemas, while `write/read` workloads only allow `blob` type as a column type.

PR #93, opened 2024-06-19
c-s: Implement `insert` operation for user profiles

muzarski:
Fix: https://github.com/scylladb/cql-stress/issues/91

This PR introduces a logic for predefined `insert` operation.

Changes:
- adjusted CLI to support `insert` operation for `user` command. Users can now choose to perform this operation (with a given sample ratio weight) via `ops()` parameter.
- implemented runtime logic for `insert` operation. As the name suggests, the operation simply generates a row and inserts it to database.
- updated README
- TODO: implement runtime tests in python

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/operation/user.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/user.rs
@@ -183,23 +183,27 @@ impl UserOperationFactory {
             "Compound partition keys are not yet supported by the tool!"
         );
 
-        let mut queries_payload = HashMap::new();
-        for (q_name, (q_def, weight)) in query_definitions {
-            queries_payload.insert(
-                q_name.to_owned(),
-                (q_def.to_prepared_statement(&session).await?, *weight),
-            );
-        }
-        // Handle 'insert' operation separately.
-        if let Some(insert_weight) = &user_profile.insert_operation_weight {
-            let insert_statement =
-                Self::prepare_insert_statement(&session, &user_profile.table, &table_metadata)
-                    .await?;
-            queries_payload.insert(
-                PREDEFINED_INSERT_OPERATION.to_owned(),
-                (insert_statement, *insert_weight),
-            );
-        }
+        let queries_payload = {
+            let mut queries_payload = HashMap::new();
+            for (q_name, (q_def, weight)) in query_definitions {
+                queries_payload.insert(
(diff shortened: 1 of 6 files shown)


muzarski:
I need to implement runtime tests for this feature, but other than that, this PR is ready for review.

src/bin/cql-stress-cassandra-stress/settings/command/user.rs
```diff
@@ -157,9 +164,10 @@ impl UserParams {
...
             mut queries,
         } = handles.profile.get().unwrap();
-        let queries_ratio = handles.ratio.get().unwrap();
+        let mut queries_ratio = handles.ratio.get().unwrap();
         let clustering: Arc<dyn DistributionFactory> = handles.clustering.get().unwrap().into();
 
+        let insert_operation_weight = queries_ratio.remove(PREDEFINED_INSERT_OPERATION);
         let queries_payload = queries_ratio
             .into_iter()
             .map(
```
wprzytula, line 173: I don't understand why this particular (key, value) pair is removed from the map here, while the remaining ones are removed later. Could you provide an example CLI input that shows examples of both this particular pair and different pairs?

wprzytula: Having read the strings later in this commit I think I understand. Can you add some comments explaining this specific treatment of `PREDEFINED_INSERT_OPERATION`?

src/bin/cql-stress-cassandra-stress/operation/user.rs
```diff
@@ -127,6 +127,31 @@ pub struct UserOperationFactory {
...
+        table_name: &str,
+        table_metadata: &Table,
+    ) -> Result<PreparedStatement> {
+        let column_names = table_metadata
+            .columns
+            .keys()
+            .map(ToOwned::to_owned)
+            .collect::<Vec<_>>();
+
+        let column_list_str = column_names.join(", ");
```
wprzytula, line 141: Nit:

wprzytula:
```suggestion
        let column_names = table_metadata
            .columns
            .keys()
            .map(String::as_str)
            .collect::<Vec<_>>();

        let column_list_str = column_names.join(", ");
```

wprzytula (approved):

PR #96, opened 2024-07-03
mixed: prepare statements only for subcommands defined by user

muzarski:
# Motivation

See the discussion: https://github.com/scylladb/cql-stress/pull/86#issuecomment-2204524664.

My last PR (https://github.com/scylladb/cql-stress/pull/86) introduced a new bug - `mixed` command tries to prepare a statement for each operation (read/write/counter_read/counter_write), even though it does not create the tables.

This PR fixes it so we prepare the statements only for the operations that are going to be sampled (were specified in `ratio()` parameter by the user). We obviously then assume that the corresponding write workload was ran beforehand and that it created tables used by the `mixed` workload.

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
+++ b/src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
@@ -19,6 +19,12 @@ impl<T: Copy> EnumeratedDistribution<T> {
     }
 }
 
+impl<T: PartialEq + Eq> EnumeratedDistribution<T> {
+    pub fn contains(&self, t: &T) -> bool {
+        self.items.iter().any(|(item, _weight)| item == t)
+    }
+}
+
 impl<T: std::fmt::Display> std::fmt::Display for EnumeratedDistribution<T> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{{")?;
--- a/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use futures::Future;
 use std::{ops::ControlFlow, sync::Arc};
 
 use cql_stress::{
@@ -26,10 +27,10 @@ use super::{
 };
 
 pub struct MixedOperation {
-    write_operation: WriteOperation,
-    counter_write_operation: CounterWriteOperation,
-    read_operation: RegularReadOperation,
-    counter_read_operation: CounterReadOperation,
+    write_operation: Option<WriteOperation>,
+    counter_write_operation: Option<CounterWriteOperation>,
+    read_operation: Option<RegularReadOperation>,
+    counter_read_operation: Option<CounterReadOperation>,
     cached_row: Option<Vec<CqlValue>>,
     workload: RowGenerator,
     max_operations: Option<u64>,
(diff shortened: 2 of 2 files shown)


src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
```diff
@@ -17,6 +17,13 @@ impl<T: Copy> EnumeratedDistribution<T> {
     pub fn sample(&self) -> T {
         self.items[self.dist.sample(&mut rand::thread_rng())].0
     }
+
+    pub fn has_item(&self, t: &T) -> bool {
+        self.items
+            .iter()
+            .map(|(item, _weight)| item)
+            .any(|item| item == t)
+    }
```
wprzytula, line 26: Nit: a customary Rust name for such method would be `contains`.

src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs
```diff
@@ -17,6 +17,13 @@ impl<T: Copy> EnumeratedDistribution<T> {
     pub fn sample(&self) -> T {
         self.items[self.dist.sample(&mut rand::thread_rng())].0
     }
+
+    pub fn has_item(&self, t: &T) -> bool {
+        self.items
+            .iter()
+            .map(|(item, _weight)| item)
+            .any(|item| item == t)
```
wprzytula, line 25: Is the `map` really needed? We can do with one `any`.

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -56,10 +56,16 @@ impl OperationFactory for MixedOperationFactory {
...
+        let write_operation = self.write_operation_factory.as_ref().map(|f| f.create());
+        let counter_write_operation = self
+            .counter_write_operation_factory
+            .as_ref()
+            .map(|f| f.create());
+        let read_operation = self.read_operation_factory.as_ref().map(|f| f.create());
+        let counter_read_operation = self
+            .counter_read_operation_factory
+            .as_ref()
+            .map(|f| f.create());
```
wprzytula, line 68: We could try to reduce boilerplate by introducing a generic helper function:
```rust
fn produce_operation<Factory: OperationFactory>(factory: &Option<Factory>) -> Factory::Product { 
    factory.as_ref().map(|f| f.create())
}

let write_operation = produce_operation(&self.counter_write_operation_factory);
...
```
I thought up the names; replace them with correct ones. I hope you get the idea.

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -88,16 +94,52 @@ impl MixedOperationFactory {
...
+                CounterReadOperationFactory::new(
+                    settings.clone(),
+                    session,
+                    DEFAULT_COUNTER_TABLE_NAME,
+                )
+                .await?,
+            )
+        } else {
+            None
+        };
```
wprzytula, line 142: WDYT about such reduction of boilerplate?
```rust
async fn conditional_create_factory<Factory: OperationFactory, Fut: Future<Output=Factory>>(kind: MixedSubcommand, f: impl FnOnce() -> Fut {
    if mixed_params.operation_ratio.has_item(&kind) {
        Some(f().await)
    } else {
        None
    }

    let read_operation_factory = conditional_crate_factory(
         MixedSubcommand::Read,
         || RegularReadOperationFactory::new(
                settings.clone(),
                session.clone(),
                DEFAULT_TABLE_NAME,
              )
    ).await;
}

File: src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs (hunk 2 above)
Comment (muzarski) [reply src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs:25]: It is not, good catch

**Comment by fruch** [comment]
I've took the PR into SCT integration test, and mixed is working as expected (i.e. doesn't fail)

**Comment by muzarski** [comment]
v2:
- addressed review comments
- moved `EnumeratedDistribution::contains` to separate impl with `PartialEq + Eq` trait bounds on `T`. Previously, we would require `T` to be `PartialEq + Eq` even for sampling methods - we obviously don't need these trait bounds to implement them.

File: src/bin/cql-stress-cassandra-stress/java_generate/distribution/enumerated.rs [hunk 5]
```diff
@@ -7,7 +7,7 @@ pub struct EnumeratedDistribution<T> {
     dist: WeightedIndex<f64>,
 }
 
-impl<T: Copy> EnumeratedDistribution<T> {
+impl<T: Copy + PartialEq + Eq> EnumeratedDistribution<T> {
     pub fn new(items: Vec<(T, f64)>) -> Result<Self> {
         let dist = WeightedIndex::new(items.iter().map(|w| w.1))?;
```
wprzytula: I can still see `PartialEq + Eq` bounds for `new`. Is this a leftover?

muzarski:
v2.1: In previous comment I mentioned that we don't need `PartialEq + Eq` trait bounds to implement sampling methods. I forgot to remove them, tho...

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -111,6 +148,18 @@ impl MixedOperationFactory {
...
+        ratios: &OperationRatio,
+        command_kind: &MixedSubcommand,
+        create_factory_fut: impl FnOnce() -> Fut,
+    ) -> Option<Factory> {
+        if ratios.contains(command_kind) {
+            Some(create_factory_fut().await)
+        } else {
+            None
+        }
+    }
```
wprzytula, line 162: In instantiations of this functions, is `Factory` equal to `Result<impl CassandraStressOperation>`? I struggled to find where this result comes from in outer code, but here it is. Maybe let's constrain the function further by imposing constraints on `Factory` to be `Result<impl CassandraStressOperation>`?

src/bin/cql-stress-cassandra-stress/operation/mixed.rs
```diff
@@ -111,6 +148,18 @@ impl MixedOperationFactory {
...
+        ratios: &OperationRatio,
+        command_kind: &MixedSubcommand,
+        create_factory_fut: impl FnOnce() -> Fut,
+    ) -> Option<Result<Factory>> {
+        if ratios.contains(command_kind) {
+            Some(create_factory_fut().await)
+        } else {
+            None
+        }
+    }
```
wprzytula, line 162: You still haven't constrained `Factory: CassandraStressOperationFactory`, but nvm.

PR #102, opened 2024-09-11
cargo: bump rust-driver version to 0.14

muzarski:
The only API breaking changes that needed to be addressed were rust driver's changes regarding query/execute API changes regarding paging.

We changed `query/execute` calls to `[query/execute]_unpaged` calls, adding the comments explaining why it makes sense to make use of unpaged queries in these places (unless it is obvious).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,9 +323,9 @@ dependencies = [
 
 [[package]]
 name = "darling"
-version = "0.20.8"
+version = "0.20.10"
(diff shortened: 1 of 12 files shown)


dkropachev (approved):
We would want to test real world driver flows, which means that we would want to use paged api for select queries.
But since at this point `cql-stress` can't even generate more than one row for partition key we can move making decision on how exactly do we do that later.

roydahan (approved):
Let's merge, relase and make sure SCT always picks the latest release.

PR #105, opened 2024-12-16
Bump 0.15.1

muzarski:


fruch:
seems like here the failure are running c-s

in #110 we switch to compare with our own fork of c-s, it might be working better...

soyacz:
closing in favor of: https://github.com/scylladb/cql-stress/pull/118

PR #110, opened 2025-03-04
feature(ci): add github actions for testing and releasing

### LINKED ISSUE #94: CI: create a workflow that creates and pushes a docker image
SCT makes use of the tool via a docker image with the tool installed.

We should create a GH actions workflow that would build the tool and install it in the docker image. The docker image should be pushed to dockerhub scylla repository.

CodeLieutenant:
Whats implemented:

- [x] CI Building and Testing (including Integration Tests with pytest)
- [x] Release process similar to Latte
- [x] Continous Integration
- [x] Fetching Cassandra-Stress from [Github](https://github.com/scylladb/cassandra-stress) instead of fetching full Apache Cassandra

Closes #94


Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+    pkg-config \
+    && cargo build --profile dist --all
+
+FROM debian:bookworm-slim AS production
+
+LABEL org.opencontainers.image.source="https://github.com/scylladb/cql-stress"
+LABEL org.opencontainers.image.title="ScyllaDB cql-stress"
+
+COPY --from=builder /usr/src/app/target/dist/cql-stress-cassandra-stress /usr/local/bin/cassandra-stress
+COPY --from=builder /usr/src/app/target/dist/cql-stress-scylla-bench /usr/local/bin/scylla-bench
```
fruch, line 21: lets leave the full names

soyacz:
@CodeLieutenant now we just need add a tag to create new release? Can you update docs around that?

CodeLieutenant:
I'll update the docs as soon as I release the first version to test the pipeline, it's really hard to test it without pushing and changing on master, as soon as I have everything green, it will be just on tag and that's it

Issue #103, opened 2024-11-06
c-s: Generate HDR histogram file during run


muzarski:
Ref: https://github.com/scylladb/cql-stress/issues/95

What needs to be done:
## CLI
Need to implement support for `-log` option. To solve this issue, only `-log hdrfile=` parameter needs to be implemented. Remaining parameters can be addressed later.

The `hdrfile=` parameter should accept the path to the output file.

## File format
Need to investigate the format of file produced by original [cassandra-stress](https://github.com/scylladb/cassandra-stress). For example, command `./cassandra-stress mixed duration=5s -rate threads=10 -pop seq=1..100000 -node 172.42.0.2 -log hdrfile=foo.log` produces the following:
```
#Logging op latencies for Cassandra Stress
#[Histogram log format version 1.3]
#[BaseTime: 1730903661.000 (seconds since epoch)]
#[StartTime: 1730903661.000 (seconds since epoch), Wed Nov 06 15:34:21 CET 2024]
"StartTimestamp","Interval_Length","Interval_Max","Interval_Compressed_Histogram"
[...]

PR #126, opened 2025-03-21
feature(hdrh_logger): added hdr histograms logger

soyacz:
Introduced hdr histogram file logger supporting different types of statement (read, write, couter_read, counter_write). Also handling `-log interval` for proper granularity.

closes: https://github.com/scylladb/cql-stress/issues/103

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/hdr_logger.rs
+++ b/src/bin/cql-stress-cassandra-stress/hdr_logger.rs
@@ -1,4 +1,4 @@
-use std::{fs::File, path::Path, time::SystemTime};
+use std::{fs::File, time::SystemTime};
 
 use crate::stats::Stats;
 use anyhow::Result;
@@ -11,24 +11,20 @@ use tokio::time::Instant;
 ///
 /// This struct manages a log writer for recording performance histograms,
 /// tracking the start time and last write time for accurate timing.
-pub struct HdrLogWriter {
-    log_writer: interval_log::IntervalLogWriter<
-        'static,
-        'static,
-        File,
-        hdrhistogram::serialization::V2Serializer,
-    >,
+pub struct HdrLogWriter<'w, 's> {
+    log_writer:
+        interval_log::IntervalLogWriter<'w, 's, File, hdrhistogram::serialization::V2Serializer>,
     start_timestamp: Instant,
     last_hdr_write: Instant,
 }
 
-impl HdrLogWriter {
-    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
-        let file = File::create(path)?;
+impl<'w, 's> HdrLogWriter<'w, 's> {
+    pub fn new(
+        file: &'w mut File,
+        serializer: &'s mut hdrhistogram::serialization::V2Serializer,
+    ) -> Result<Self> {
         let start_time = SystemTime::now();
 
-        let serializer = Box::new(hdrhistogram::serialization::V2Serializer::new());
-        // Note: Box::leak is used to satisfy 'static lifetime requirements of IntervalLogWriter.
         let log_writer = interval_log::IntervalLogWriterBuilder::new()
             .add_comment(
                 format!(
@@ -40,8 +36,7 @@ impl HdrLogWriter {
             .with_start_time(start_time)
             .with_base_time(start_time)
             .with_max_value_divisor(1000000.0)
-            .begin_log_with(Box::leak(Box::new(file)), Box::leak(serializer))
-            .unwrap();
+            .begin_log_with(file, serializer)?;
 
         Ok(Self {
             log_writer,
--- a/src/bin/cql-stress-cassandra-stress/hdr_logger.rs
+++ b/src/bin/cql-stress-cassandra-stress/hdr_logger.rs
@@ -1,0 +1,79 @@
+use std::{fs::File, path::Path, time::SystemTime};
+
+use crate::stats::Stats;
+use anyhow::Result;
+use cql_stress::version::get_version_info;
+use hdrhistogram::serialization::interval_log;
+use hdrhistogram::Histogram;
+use tokio::time::Instant;
+
+/// Writes histogram data to a file using HDR format.
+///
+/// This struct manages a log writer for recording performance histograms,
+/// tracking the start time and last write time for accurate timing.
+pub struct HdrLogWriter {
+    log_writer: interval_log::IntervalLogWriter<
+        'static,
+        'static,
+        File,
+        hdrhistogram::serialization::V2Serializer,
+    >,
+    start_timestamp: Instant,
+    last_hdr_write: Instant,
+}
+
+impl HdrLogWriter {
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let file = File::create(path)?;
+        let start_time = SystemTime::now();
+
+        let serializer = Box::new(hdrhistogram::serialization::V2Serializer::new());
+        // Note: Box::leak is used to satisfy 'static lifetime requirements of IntervalLogWriter.
+        let log_writer = interval_log::IntervalLogWriterBuilder::new()
+            .add_comment(
+                format!(
+                    "[Logged with Cql-stress {}]",
+                    get_version_info().cql_stress_version
+                )
+                .as_str(),
+            )
+            .with_start_time(start_time)
+            .with_base_time(start_time)
+            .with_max_value_divisor(1000000.0)
+            .begin_log_with(Box::leak(Box::new(file)), Box::leak(serializer))
+            .unwrap();
+
+        Ok(Self {
+            log_writer,
+            start_timestamp: Instant::now(),
+            last_hdr_write: Instant::now(),
+        })
+    }
+
+    /// Writes combined statistics histograms to the HDR log file.
+    ///
+    /// # Arguments
+    /// * `partial_stats` - The statistics containing histograms to log.
+    ///
+    /// # Errors
+    /// Returns an error if writing to the log file fails.
+    pub fn write_to_hdr_log(&mut self, partial_stats: &Stats) -> Result<()> {
+        let duration = self.last_hdr_write.elapsed();
+        let elapsed = self.start_timestamp.elapsed();
+
+        let mut tag_histograms: Vec<(&String, &Histogram<u64>)> =
+            partial_stats.get_histograms().iter().collect();
+        tag_histograms.sort_by(|(tag_a, _), (tag_b, _)| tag_a.cmp(tag_b));
+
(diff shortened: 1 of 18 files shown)


src/bin/cql-stress-cassandra-stress/operation/write.rs
```diff
@@ -43,6 +43,10 @@ impl CassandraStressOperation for WriteOperation {
     fn generate_row(&self, row_generator: &mut RowGenerator) -> Vec<CqlValue> {
         row_generator.generate_row()
     }
+
+    fn operation_tag(&self) -> &'static str {
+        "write"
```
vponomaryov, line 48: Same here as in the `src/bin/cql-stress-cassandra-stress/operation/mod.rs` module.
It doesn't match CS, moreover, CS has 2 variants, not 1 - for coordinated omission fixed case and for the affected.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -293,6 +299,10 @@ impl RowValidator for EqualRowValidator {
         );
         Ok(())
     }
+
+    fn operation_tag() -> &'static str {
+        "read"
```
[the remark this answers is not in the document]
soyacz: you mean just changing tags names depending on throttling param?

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
```diff
@@ -0,0 +1,74 @@
...
+                interval = interval_str
+                    .parse::<u64>()
+                    .map_err(|_| anyhow::anyhow!("Invalid interval value: {:?}", interval_str))?;
+            } else {
+                return Err(anyhow::anyhow!("Unknown parameter: {:?}", arg));
+            }
+        }
+
+        Ok(Self { hdr_file, interval })
+    }
```
muzarski, line 65: I'd rewrite the parameter parsing using `ParamsParser` that I implemented. You can look at examples in other `.rs` files in `option` module.

The most important thing is to prepare the parser. (See `prepare_parser` functions in other modules). For this one, the definition of parameters would look like:
```rust
fn prepare_parser() -> ... {
    let mut parser = ParamsParser::new(LogOption::CLI_STRING);
    
    let interval: SimpleParamHandle<IntervalMillisOrSeconds> = ParamsParser::new("interval=", Some("1s"), "some description", false);
    let hdrfile: SimpleParamHandle<String> = parser.simple_param("hdrfile=", None, "some description", false);
   
   // Group the parameters based on original cassandra-stress help.
   // Return the parser and handles to the parameters.
   ...
}
```

`IntervalMillisOrSeconds` is the type you can define and implement `Parsable` for it in `settings/param/types.rs`. Again, you can look at other examples. Feel free to ask questions if something is not clear :).

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
```diff
@@ -0,0 +1,74 @@
...
+        let mut interval = 1;
+
+        for arg in args {
+            if let Some(path) = arg.strip_prefix("hdrfile=") {
+                let path = PathBuf::from(path);
+                hdr_file = Some(path);
+            } else if let Some(interval_str) = arg.strip_prefix("interval=") {
+                interval = interval_str
+                    .parse::<u64>()
+                    .map_err(|_| anyhow::anyhow!("Invalid interval value: {:?}", interval_str))?;
```
muzarski, line 58: When I provide some garbage value to `interval=foo` in original c-s, it says:
```
Invalid value foo; must match pattern [0-9]+(ms|s|)
```

It means that current approach is not consistent with original c-s. This is why I suggest to define the `IntervalMillisOrSeconds` type so we parse this parameter the same way original c-s does. See my other comment on where to define such type. It should implement `Parsable` trait. The `Parsable::Parsed` type would probably be just `std::time::Duration` in such case.

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
```diff
@@ -0,0 +1,74 @@
...
+        "Specify logging options"
+    }
+
+    pub fn print_help() {
+        println!("Usage: {} <param list>", Self::CLI_STRING);
+        println!("    hdrfile=<filename>   - Log HDR Histogram data to the specified file");
+        println!(
+            "    interval=<seconds>   - Set the interval between logs in seconds (default: 1)"
+        );
+    }
```
muzarski, line 32: Notice that `ParamsParser` exposes `print_help()` method. See my other comment on how to create a `ParamsParser` instance.

muzarski (changes requested):

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 edition = "2021"
 name = "cql-stress"
-version = "0.1.0"
+version = "0.2.0"
```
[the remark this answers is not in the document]
soyacz: removed, I'll introduce it when we want to release

src/bin/cql-stress-cassandra-stress/operation/mod.rs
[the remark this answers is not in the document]
soyacz: ok, bit more complex, but I hope I handled it correctly: adding proper suffix in `account_operation`

src/bin/cql-stress-cassandra-stress/operation/write.rs
soyacz: resolved

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
soyacz: done

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
soyacz: done

soyacz:
@muzarski @vponomaryov thanks for your feedback, I think I resolved your comments. Please re-review.

vponomaryov (changes requested):
<deleted>

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    #[test]
+    fn log_bad_interval_test() {
+        let args = vec!["interval=foo"];
+        let (parser, _) = prepare_parser();
+
+        // Should fail with an invalid interval format
+        assert!(parser.parse(args).is_err());
+    }
+}
```
muzarski, line 147: Thanks for the tests! You can also update `cs_args_[good/bad]_test.in` files. More tests never hurt us.

muzarski (changes requested):
Last iteration of review - apart from that, looks really good. Nice job.

src/bin/cql-stress-cassandra-stress/hdr_logger.rs
```diff
@@ -0,0 +1,79 @@
...
+            .with_max_value_divisor(1000000.0)
+            .begin_log_with(Box::leak(Box::new(file)), Box::leak(serializer))
+            .unwrap();
+
+        Ok(Self {
+            log_writer,
+            start_timestamp: Instant::now(),
+            last_hdr_write: Instant::now(),
+        })
+    }
```
muzarski, line 51: 1. no need to leak
2. why `unwrap()` when we return `Result<>` already

I've prepared a commit that fixes these two issues: https://github.com/muzarski/cql-stress/commit/6cf10bc36fa1d5f928c47fa80de72e37f6d80d13. It also removes the `panic!()` from `main.rs` (see my other comment)

muzarski (changes requested):
Sorry, for some reason GH decided not to include the most important comment when I submitted the review last time.

src/bin/cql-stress-cassandra-stress/stats.rs
```diff
@@ -76,25 +113,60 @@ impl sharded_stats::StatsFactory for StatsFactory {
...
-                    .unwrap();
+                let metrics = self.latency_calculator.calculate(ctx);
+                let default_latency = self.latency_calculator.default_latency(&metrics);
+                self.latency_histogram.record(default_latency).unwrap();
+
+                let service_time_tag = format!("{}-st", tag);
+                let service_time_histogram = self
+                    .histograms
+                    .entry(service_time_tag)
+                    .or_insert_with(|| Histogram::new(3).unwrap());
```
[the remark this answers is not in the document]
soyacz: It's histogram precision. used HISTOGRAM_PRECISION const for that which should be self explanatory I think

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -62,6 +62,9 @@ pub trait CassandraStressOperation: Sync + Send {
 
     fn execute(&self, row: &[CqlValue]) -> impl Future<Output = Result<ControlFlow<()>>> + Send;
     fn generate_row(&self, row_generator: &mut RowGenerator) -> Vec<CqlValue>;
+    fn operation_tag(&self) -> &'static str {
+        ""
+    }
```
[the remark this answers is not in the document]
soyacz: IIRC it should return name from user profile. I tried implementing it but I got into problems with 'static' lifetime of operation_tag. I need to think/learn how to fix that.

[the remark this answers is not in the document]
muzarski: Yeah, the `'static` constraint should probably be ditched. We get to know the name from user profile only during runtime, thus the lifetime of this `str` won't be `'static` for sure. You could probably relax the constraints, and simply return `&str`.

src/bin/cql-stress-cassandra-stress/stats.rs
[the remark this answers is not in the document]
muzarski: Ok, fair enough

src/bin/cql-stress-cassandra-stress/hdr_logger.rs
soyacz: awesome! applied your change

src/bin/cql-stress-cassandra-stress/settings/option/log.rs
soyacz: added

src/bin/cql-stress-cassandra-stress/operation/mod.rs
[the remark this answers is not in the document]
soyacz: indeed it was redundant and I made user profiles also to be hdr-logged properly

muzarski (approved):
LGTM!

Issue #108, opened 2025-01-28
`-rate` API is not compatible with the C-S original API


vponomaryov:
The `cql-stress-cassandra-stress` is designed to be drop-in replacement.

Expected C-S command: `cassandra-stress write no-warmup cl=QUORUM duration=5s -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate 'threads=3 fixed=400/s' -node 172.17.0.4`

Actual results:
```
root@986b5bd98b2c:/# cql-stress-cassandra-stress write no-warmup cl=QUORUM duration=5s -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate 'threads=3 fixed=400/s' -node 172.17.0.4
Usage:      cassandra-stress <command> [options]
Help usage: cassandra-stress help <command|option>

---Commands---
help                 : Print help for a command or option
write                : Multiple concurrent writes against the cluster.
read                 : Multiple concurrent reads - the cluster must first be populated by a write test.
counter_write        : Multiple concurrent updates of counters.
counter_read         : Multiple concurrent reads of counters. The cluster must first be populated by a counterwrite test.
mixed                : Interleaving of any basic commands, with configurable ratio and distribution - the cluster must first be populated by a write test.
user                 : Interleaving of user provided queries, with configurable ratio and distribution - the cluster must first be populated by a write test.

---Options---
-node                : Nodes to connect to
-rate                : Thread count, rate limit or automatic mode (default is auto)
-mode                : CQL connection options
-schema              : Replication settings, compression, compaction, etc.
-col                 : Column details such as size distribution, names
-pop                 : Population distribution

Failed to parse parameter fixed.

Caused by:
    Invalid value =400/s. Boolean flag cannot have any value.
Error: Failed to parse CLI arguments.
```

It turns out that to achieve the same goal we need to change the command to the following variant:
`cql-stress-cassandra-stress write no-warmup cl=QUORUM duration=5s -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate 'threads=3 throttle=400/s fixed' -node 172.17.0.4`

[...]

PR #127, opened 2025-03-28
fix(rate): adjusted -rate option to match c-s

soyacz:
Seems `-rate` fixed option changed in c-s and this code reflects that change:
 if `fixed=op/s` is set, it enables `co_fixed` flag and when `throttle=op/s` is used `co_fixed` flag is `false`

fixes: https://github.com/scylladb/cql-stress/issues/108

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
@@ -76,7 +76,7 @@ impl RateOption {
     fn from_handles(handles: RateParamHandles) -> Result<Self> {
         let threads = handles.threads.get();
         let throttle = handles.throttle.get();
-        let co_fixed = handles.co_fixed.get().is_some();
+        let fixed = handles.fixed.get();
         let min_threads = handles.threads_gte.get();
         let max_threads = handles.threads_lte.get();
         let auto = handles.auto.get().is_some();
@@ -87,13 +87,21 @@ impl RateOption {
                 max_threads,
                 auto,
             },
-            _ => ThreadsInfo::Fixed {
-                // SAFETY: The parameters are grouped in a way that this won't ever panic
-                // when entering this branch.
-                threads: threads.unwrap(),
-                throttle,
-                co_fixed,
-            },
+            _ => {
+                let (final_throttle, co_fixed) = match (fixed, throttle) {
+                    (Some(rate), None) => (Some(rate), true),
+                    (None, Some(rate)) => (Some(rate), false),
+                    (None, None) => (None, false),
+                    (Some(_), Some(_)) => {
+                        return Err(anyhow::anyhow!("Cannot specify both fixed and throttle"));
+                    }
+                };
+                ThreadsInfo::Fixed {
+                    threads: threads.unwrap(),
+                    throttle: final_throttle,
+                    co_fixed,
+                }
+            }
         };
 
         Ok(Self { threads_info })
@@ -103,7 +111,7 @@ impl RateOption {
 struct RateParamHandles {
     pub threads: SimpleParamHandle<u64>,
     pub throttle: SimpleParamHandle<Rate>,
-    pub co_fixed: SimpleParamHandle<bool>,
+    pub fixed: SimpleParamHandle<Rate>,
     pub threads_gte: SimpleParamHandle<u64>,
     pub threads_lte: SimpleParamHandle<u64>,
     pub auto: SimpleParamHandle<bool>,
@@ -116,13 +124,13 @@ fn prepare_parser() -> (ParamsParser, RateParamHandles) {
     let throttle = parser.simple_param(
         "throttle=",
         None,
-        "throttle operations per second across all clients to a maximum rate (or less) with no implied schedule",
+        "limits the rate of operations per second with a schedule; see `fixed` option if latency adjustment to coordinated-omission is needed",
         false,
     );
-    let co_fixed = parser.simple_param(
-        "fixed",
+    let fixed = parser.simple_param(
+        "fixed=",
         None,
-        "display coordinated-omission-fixed latencies",
+        "limits the rate of operations per second with a schedule; displays coordinated-omission-fixed latencies (i.e. end_time - scheduled_start_time); for more information see https://www.scylladb.com/2021/04/22/on-coordinated-omission/",
         false,
     );
     let threads_gte = parser.simple_param(
@@ -145,10 +153,10 @@ fn prepare_parser() -> (ParamsParser, RateParamHandles) {
     );
 
     // $ ./cassandra-stress help -rate
-    // Usage: -rate threads=? [throttle=?] [fixed]
+    // Usage: -rate threads=? [throttle=?] [fixed=?]
     //  OR
     // Usage: -rate [threads>=?] [threads<=?] [auto]
-    parser.group(&[&threads, &throttle, &co_fixed]);
+    parser.group(&[&threads, &throttle, &fixed]);
     parser.group(&[&threads_gte, &threads_lte, &auto]);
 
     (
@@ -156,7 +164,7 @@ fn prepare_parser() -> (ParamsParser, RateParamHandles) {
         RateParamHandles {
             threads,
             throttle,
-            co_fixed,
+            fixed,
             threads_gte,
             threads_lte,
             auto,
@@ -171,7 +179,25 @@ mod tests {
     use super::prepare_parser;
 
     #[test]
-    fn rate_good_params_group_one_test() {
+    fn rate_with_fixed_test() {
+        let args = vec!["threads=100", "fixed=15/s"];
+        let (parser, handles) = prepare_parser();
+
+        assert!(parser.parse(args).is_ok());
+
+        let params = RateOption::from_handles(handles).unwrap();
+        assert_eq!(
+            ThreadsInfo::Fixed {
+                threads: 100,
+                throttle: Some(15),
+                co_fixed: true
+            },
+            params.threads_info
+        );
+    }
+
+    #[test]
+    fn rate_with_throttle_test() {
         let args = vec!["threads=100", "throttle=15/s"];
         let (parser, handles) = prepare_parser();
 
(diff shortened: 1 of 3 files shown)


src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
```diff
@@ -119,10 +127,10 @@ fn prepare_parser() -> (ParamsParser, RateParamHandles) {
         "throttle operations per second across all clients to a maximum rate (or less) with no implied schedule",
         false,
     );
-    let co_fixed = parser.simple_param(
-        "fixed",
+    let fixed = parser.simple_param(
+        "fixed=",
         None,
-        "display coordinated-omission-fixed latencies",
+        "expect fixed rate of operations per second across all clients with implied schedule",
```
vponomaryov, line 133: `co_fixed` is more correct naming.
It means `Coordinated Omission FIXED latencies`.

So, 2 concerns here.
At first, I think it is more correct to use exactly `co_fixed` all over the code and keep `fixed` only for user-facing param name due to the compatibility with the C-S.
Second, the `expect fixed rate of operations per second across all clients with implied schedule` statement is correct for both  - `throttle` and `co_fixed` variants.
They differ ONLY in latency calculation.
They behave completely the same from the operations flow point of view.

Also, knowing the HDR histograms tags behavior, the previous statement `display coordinated-omission-fixed latencies` was precise for 100%.
We always show `-st` tags (coordinated omission affected) ones and print `-rt` ones (coordinated omission fixed) only when `fixed` (`co_fixed`) param is specified.

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
```diff
@@ -87,13 +87,21 @@ impl RateOption {
...
             },
-            _ => ThreadsInfo::Fixed {
-                // SAFETY: The parameters are grouped in a way that this won't ever panic
-                // when entering this branch.
-                threads: threads.unwrap(),
-                throttle,
-                co_fixed,
-            },
+            _ => {
+                let (final_throttle, co_fixed) = match (fixed_rate, throttle) {
```
vponomaryov, line 91: it is not `fixed_rate`, it is `fixed coordinated omission latencies when rate is specified`.
`throttle` here is `coordinated omission affected latencies when rate is specified.`

See https://www.scylladb.com/2021/04/22/on-coordinated-omission/

vponomaryov (changes requested):

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
soyacz: all over the code we use `ThreadsInfo::Fixed` enum with `Fixed` having `co_fixed` attr - so all over the code we use it. I renamed here `co_fixed` to `fixed` as user now is specifying fixed rate in op/s instead of setting a flag.

As to message shown to user, it's exactly the same as in c-s - this is why I changed it to follow c-s approach. Current message didn't fit anymore - again, we don't control display only but now specify fixed rate of op/s meaning we care not only about throttle but if system keeps up with this value.

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
soyacz: I still think it's fixed rate: user defines if cql-stress should just throttle to specified value or if is trying to keep fixed rate and show increased latencies if system can't keep up (taking into account delay in sending request from schedule).

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
vponomaryov: > ... I renamed here `co_fixed` to `fixed` as user now is specifying fixed rate in op/s instead of setting a flag.

Once again, "fixed" is not rate, but the way how latencies get calculated.

> As to message shown to user, it's exactly the same as in c-s - this is why I changed it to follow c-s approach. Current message didn't fit anymore - again, we don't control display only but now specify fixed rate of op/s meaning we care not only about throttle but if system keeps up with this value.

`but now specify fixed rate of op/s ` - false, the word "fixed" here means only "corrected way of calculating latencies", not "making the rate static".

Please, read https://www.scylladb.com/2021/04/22/on-coordinated-omission/

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
vponomaryov: Look at the request scheduler in the code if you don't believe me.
Request flow is completely the same for both - `throttle` and `fixed` (`co_fixed`) parameters.
Latencies calculation is `the only` difference and reason for having two parameters instead of one.

soyacz: still we are discussing user facing part, while internals didn't change, I don't understand the problem

vponomaryov: `fixed_rate` is missleading, because we have the same static number of op/s in both cases.

soyacz: ok, I don't mind changing this local var name, changed to fixed

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
soyacz: My understanding is: throttle - we don't care if it's less, just don't exceed this value, while `fixed` - we care if system can sustain provided rate and if not, show latencies that take into account latency seen from user perspective (from the moment statement was planned to be scheduled). I'll read this doc later, it's quite long.

Doc's and behavior now matches cassandra-stress to fix [issue](https://github.com/scylladb/cql-stress/issues/108)
what do you propose in that situation?

vponomaryov: > My understanding is: throttle - we don't care if it's less, just don't exceed this value, while `fixed` - we care if system can sustain provided rate and if not, ...

It is NOT what is coded under the hood. Neither in C-S, nor in cql-stress.

> Doc's and behavior now matches cassandra-stress to fix [issue](https://github.com/scylladb/cql-stress/issues/108)

This issue is just user interface compatibility issue having idea of the "drop-in" replacement.
It has smaller severity than messing up with the coordinated omission handling code and docs for it.

> what do you propose in that situation?

I propose to study the `coordinated omission phenomenon and how it is handled in C-S and cql-stress and only then continue with this change`.

soyacz: ok, I can change help message. Any proposals?

as to deep code changes - it's not true, variable name was changed only in `RateParamHandles` which is not pub and used only within this file for parsing. I changed it as it fixed now changed meaning too. Other underlying logic behind load generation and latency measurement didn't change.

vponomaryov: > > ok, I can change help message. Any proposals?
> 
> First of all, I believe that `throttle=` help message is incorrect. It mentions **with no implied schedule**, while we clearly have a `RateLimiter` struct that schedules the operations - it computes the `scheduled_start_time` for each operation which may be different than the `actual_start_time`. Am I correct on that, @vponomaryov ?

Correct.
The observable diff appears when DB cluster cannot sustain the `requested` load (`-rate ... n op/s`).
When we do not specify the `-rate` we do not have schedule and no coordinated omission phenomenon takes place in this case.

> So I think it should be adjusted as well. My suggestions:
> 
> * `throttle` - limits the rate of operations per second with a schedule; does not fix the displayed latencies in regards to coordinated-omission problem (i.e. end_time - actual_start_time)
> * `fixed` - limits the rate of operations per second with a schedule; displays coordinated-omission-fixed latencies (i.e. end_time - scheduled_start_time)

+

soyacz: thanks @muzarski and @vponomaryov I've set to your proposal

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
```diff
@@ -116,13 +124,13 @@ fn prepare_parser() -> (ParamsParser, RateParamHandles) {
     let throttle = parser.simple_param(
         "throttle=",
         None,
-        "throttle operations per second across all clients to a maximum rate (or less) with no implied schedule",
+        "limits the rate of operations per second with a schedule; does not fix the displayed latencies in regards to coordinated-omission problem (i.e. end_time - actual_start_time)",
```
fruch, line 127:
```suggestion
        "limits the rate of operations per second with a schedule; see `fixed` option if coordinated-omission is needed",
```

it's a bit weird to me, to express with such details what it's not gonna do :)
but just point to where is can be explain with more

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
```diff
@@ -116,13 +124,13 @@ fn prepare_parser() -> (ParamsParser, RateParamHandles) {
...
         false,
     );
-    let co_fixed = parser.simple_param(
-        "fixed",
+    let fixed = parser.simple_param(
+        "fixed=",
         None,
-        "display coordinated-omission-fixed latencies",
+        "limits the rate of operations per second with a schedule; displays coordinated-omission-fixed latencies (i.e. end_time - scheduled_start_time)",
```
fruch, line 133: maybe putting a link to the blog ?

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
muzarski: Thanks, I wasn't sure how to phrase it properly. But I'd still be more specific: ``see `fixed` option if latency adjustment to coordinated-omission is needed"``. Coordinated-omission is not needed - what user might need is the latency correction if C-O appears.

muzarski (approved):
I agree with @fruch suggestions. Apart from that, LGTM.

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
soyacz: I also was thinking about that. Added.

src/bin/cql-stress-cassandra-stress/settings/option/rate.rs
soyacz: fixed

vponomaryov (approved):
LGTM

Issue #136, opened 2025-04-18
add support for -transport option (support client  encryption)


soyacz:
`-transport` option allows using encrypted communication between client-scylla

fruch:
@soyacz can you reference which SCT test case you need it most ?

soyacz:
sanity test: [test-cases/longevity/longevity-100gb-4h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/master/test-cases/longevity/longevity-100gb-4h.yaml) sets client encrypt to true

soyacz:
created PR: https://github.com/scylladb/cql-stress/pull/137

PR #137, opened 2025-04-23
feature(c-s): add -transport option support

soyacz:
Introducing basic `-transport` support to cassandra-stress frontend.

- [x] - [test with dev version](https://argus.scylladb.com/tests/scylla-cluster-tests/0cd19b7e-fb77-4661-89b6-b17d6c812db9)

closes: https://github.com/scylladb/cql-stress/issues/136

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/main.rs
+++ b/src/bin/cql-stress-cassandra-stress/main.rs
@@ -22,6 +22,7 @@ use cql_stress::{
     sharded_stats::StatsFactory as _,
 };
 use hdr_logger::HdrLogWriter;
+
 #[cfg(feature = "user-profile")]
 use operation::UserOperationFactory;
 use operation::{
@@ -156,6 +157,11 @@ async fn prepare_run(
 
     if let Some(creds) = &settings.mode.user_credentials {
         builder = builder.user(&creds.username, &creds.password);
+    }
+
+    if settings.transport.truststore.is_some() || settings.transport.keystore.is_some() {
+        let ssl_ctx = settings.transport.generate_ssl_context()?;
+        builder = builder.tls_context(Some(ssl_ctx));
     }
 
     let default_exec_profile = ExecutionProfile::builder()
--- a/src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
@@ -1,0 +1,204 @@
+use crate::settings::param::types::{FlagNumericOrBool, NonEmptyString, NotSupported};
+use crate::settings::param::{ParamsParser, SimpleParamHandle};
+use anyhow::Result;
+use std::collections::HashMap;
+
+#[derive(Debug, Clone, Default)]
+pub struct TransportOption {
+    pub factory: Option<NotSupported>,
+    pub truststore: Option<String>,
+    pub truststore_password: Option<NotSupported>,
+    pub keystore: Option<String>,
+    pub keystore_password: Option<NotSupported>,
+    pub ssl_protocol: Option<NotSupported>,
+    pub ssl_alg: Option<NotSupported>,
+    pub store_type: Option<NotSupported>,
+    pub ssl_ciphers: Option<NotSupported>,
+    pub hostname_verification: Option<bool>,
+}
+
+struct TransportParamHandles {
+    factory: SimpleParamHandle<NotSupported>,
+    truststore: SimpleParamHandle<NonEmptyString>,
+    truststore_password: SimpleParamHandle<NotSupported>,
+    keystore: SimpleParamHandle<NonEmptyString>,
+    keystore_password: SimpleParamHandle<NotSupported>,
+    ssl_protocol: SimpleParamHandle<NotSupported>,
+    ssl_alg: SimpleParamHandle<NotSupported>,
+    store_type: SimpleParamHandle<NotSupported>,
+    ssl_ciphers: SimpleParamHandle<NotSupported>,
+    hostname_verification: SimpleParamHandle<FlagNumericOrBool>,
+}
+
+fn prepare_parser() -> (ParamsParser, TransportParamHandles) {
+    let mut parser = ParamsParser::new(TransportOption::CLI_STRING);
+    let factory = parser.simple_param("factory=", None, "SSL factory class (unsupported)", false);
+    let truststore = parser.simple_param("truststore=", None, "Path to truststore file", false);
+    let truststore_password = parser.simple_param(
+        "truststore-password=",
+        None,
+        "Truststore password (unsupported)",
+        false,
+    );
+    let keystore = parser.simple_param("keystore=", None, "Path to keystore file", false);
+    let keystore_password = parser.simple_param(
+        "keystore-password=",
+        None,
+        "Keystore password (unsupported)",
+        false,
+    );
+    let ssl_protocol =
+        parser.simple_param("ssl-protocol=", None, "SSL protocol (unsupported)", false);
+    let ssl_alg = parser.simple_param("ssl-alg=", None, "SSL algorithm (unsupported)", false);
+    let store_type = parser.simple_param("store-type=", None, "Store type (unsupported)", false);
+    let ssl_ciphers = parser.simple_param("ssl-ciphers=", None, "SSL ciphers (unsupported)", false);
+    let hostname_verification = parser.simple_param(
+        "hostname-verification=",
+        None,
+        "Enable hostname verification (true/false/1/0)",
+        false,
+    );
+    parser.group(&[
+        &factory,
+        &truststore,
+        &truststore_password,
+        &keystore,
+        &keystore_password,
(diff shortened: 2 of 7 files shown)


CodeLieutenant (approved):

muzarski:
> @muzarski can you take a look when you find some time?

Yes, sorry for the wait. I'll review it tomorrow.

src/bin/cql-stress-cassandra-stress/main.rs
```diff
@@ -158,6 +160,46 @@ async fn prepare_run(
...
+                    format!("Failed to set certificate file: {}", key_path.display())
+                })?;
+            builder
+                .set_private_key_file(&key_path, SslFiletype::PEM)
+                .with_context(|| {
+                    format!("Failed to set private key file: {}", key_path.display())
+                })?;
+        }
+        Ok(builder.build())
+    }
```
muzarski, line 196: Maybe it should be a method of `TransportOption`?

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,212 @@
...
+            ssl_ciphers,
+            hostname_verification,
+        },
+    )
+}
+impl TransportOption {
+    pub const CLI_STRING: &'static str = "-transport";
+
+    pub fn parse(payload: &mut HashMap<String, Vec<&str>>) -> Result<Self> {
+        fn from_handles(handles: TransportParamHandles) -> Result<TransportOption> {
```
muzarski, line 92: Let's make this a method on `TransportOption` instead of a nested function.

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,212 @@
...
+            let mut opt = TransportOption::default();
```
muzarski, line 93: Why not construct `TransportOption` object at the end? It's not Rusty way to create a mutable object, and only then initialize its fields.

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,212 @@
...
+            if let Some(factory) = handles.factory.get() {
+                if factory.is_empty() {
+                    return Err(anyhow::anyhow!("Empty value for factory"));
+                }
+                opt.factory = Some(factory);
+            }
```
muzarski, line 99:
```rust
let factory = handles.factory.get()
         .map(|f| if factory.is_empty() { Err(...) } 
                         else { Ok(f) })
         .transpose()?; 
```

The same for remaining fields. But actually, you could introduce a `Parsable` type, such as `NonEmptyString`. Maybe there is already one - I don't remember. See `settings/param/types.rs`.

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,212 @@
...
+                    "false" | "0" => opt.hostname_verification = Some(false),
+                    "" => {}
+                    _ => {
+                        return Err(anyhow::anyhow!(format!(
+                            "Invalid value for hostname-verification: {}",
+                            hv
+                        )));
+                    }
+                }
+            }
```
muzarski, line 160: Again, this logic could be implemented using `Parsable` trait. Create a type (e.g. `FlagNumericOrBool`) and implement `Parsable` for it..

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,212 @@
...
+            if let Some(keystore) = handles.keystore.get() {
+                if keystore.is_empty() {
+                    return Err(anyhow::anyhow!("Empty value for keystore"));
+                }
+                opt.keystore = Some(keystore);
+            }
+            if let Some(ks_pass) = handles.keystore_password.get() {
+                if !ks_pass.is_empty() {
+                    println!("[WARN] -transport keystore-password is set but not supported");
```
muzarski, line 120: Maybe it makes sense to use `log::warn!` or `tracing::warn!` (I don't remember which one is used in cql-stress).

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,212 @@
...
+            println!("  ssl-ciphers: {}", v);
+        }
+        if let Some(v) = self.hostname_verification {
+            println!("  hostname-verification: {}", v);
+        }
+    }
+
+    pub fn description() -> &'static str {
+        "transport and SSL options: [factory=?] [truststore=?] [truststore-password=?] [keystore=?] [keystore-password=?] [ssl-protocol=?] [ssl-alg=?] [store-type=?] [ssl-ciphers=?] [hostname-verification=?]"
+    }
```
muzarski, line 206: I think `transport and SSL options` is just fine. The parameters should be displayed only when user types `c-s help -transport`, and this is already handled by `print_help` method.

muzarski (changes requested):

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
soyacz: moved

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
soyacz: fixed

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
soyacz: I added `NonEmptyString` type

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
soyacz: added `FlagNumericOrBool` type

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
soyacz: I didn't know - there were one usage of `tracing::warn!` - so I used this approach. I also added a new type for it `NotSupported` so the logic is reused.

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
soyacz: fixed

soyacz:
@muzarski great thanks for your review. I made recommended/requested changes and is ready for re-review.

src/bin/cql-stress-cassandra-stress/settings/option/transport.rs
```diff
@@ -0,0 +1,231 @@
...
+        let (parser, handles) = prepare_parser();
+        parser.parse(params)?;
+        TransportOption::from_handles(handles)
+    }
+
+    fn from_handles(handles: TransportParamHandles) -> Result<TransportOption> {
+        let factory = match handles.factory.get() {
+            Some(f) => Some(NonEmptyString::parse(&f)?.0),
+            None => None,
+        };
```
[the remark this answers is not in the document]
muzarski: You don't have to `match`, but most importantly - you don't have to explicitly call `NonEmptyString::parse` here.

soyacz:
@muzarski I refactored according to your comments. Please check out the last commit (I'll squash them before merging)

muzarski:
> @muzarski I refactored according to your comments. Please check out the last commit (I'll squash them before merging)

Nice, looks much cleaner - thanks to that I realized just how many of the parameters are unsupported - quite a lot.

I'm not sure if representing unsupported type as `()` is a good pattern, but I don't have a better idea. Maybe:
```rust
pub struct NotSupported;

impl Parsable for NotSupported {
    type Parsed = NotSupported;
    fn parse(s: &str) -> Result<Self::Parsed> {
        tracing::warn!("This parameter is not supported and will be ignored: {}", s);
        Ok(())
    }
}
```

And then representing the fields as `NotSupported `instead of `()`? I'm not sure - I'll let you decide :).

Please, address the failing test, clippy lints and I think we can merge.

soyacz:
> Nice, looks much cleaner - thanks to that I realized just how many of the parameters are unsupported - quite a lot.

I implemented only what was required in our testing. We'll probably support more when needed.
> 
> I'm not sure if representing unsupported type as `()` is a good pattern, but I don't have a better idea. Maybe:
> 
> ```rust
> pub struct NotSupported;
> 
> impl Parsable for NotSupported {
>     type Parsed = NotSupported;
>     fn parse(s: &str) -> Result<Self::Parsed> {
>         tracing::warn!("This parameter is not supported and will be ignored: {}", s);
>         Ok(())
>     }
> }
> ```
> 
> And then representing the fields as `NotSupported `instead of `()`? I'm not sure - I'll let you decide :).
> 
I decided to use NotSupported

muzarski (approved):

Issue #138, opened 2025-05-05
Validation errors after adding column to table


soyacz:
## Packages

Scylla version: `2025.2.0~dev-20250424.76cd707b18e2` with build-id `65a34ed23af5208f45e783a08889483c2944cd92`

Kernel Version: `6.8.0-1027-aws`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

disrupt_add_remove_column causes cql-stress to fail validation due to result row containing additional `None` values for added columns:
```
error=Expected row's [Blob([49, 76, 75, 52, 75, 50, 57, 50, 57, 48]), Blob([...])]) length: 2
Result row's ([Some(Blob([49, 76, 75, 52, 75, 50, 57, 50, 57, 48])), Some(Blob([...])), None, None, None]) length: 5 
partition_key=Blob([49, 76, 75, 52, 75, 50, 57, 50, 57, 48])

```
Basically, cql-stress should not query for `*` columns - need to specify them explicitly so we don't receive values for added ones.

## Impact

Failing test critically

## How frequently does it reproduce?

Always when table get's new columns

## Installation details

Cluster size: 6 nodes (i4i.2xlarge)

Scylla Nodes used in this run:
- longevity-50gb-12h-1876-cql-db-node-c0385ef4-6 (46.137.50.171 | 10.4.1.104) (shards: 6)
- longevity-50gb-12h-1876-cql-db-node-c0385ef4-5 (176.34.67.195 | 10.4.1.107) (shards: 6)
- longevity-50gb-12h-1876-cql-db-node-c0385ef4-4 (54.247.207.188 | 10.4.1.196) (shards: 6)
- longevity-50gb-12h-1876-cql-db-node-c0385ef4-3 (3.250.75.254 | 10.4.1.147) (shards: 4)
- longevity-50gb-12h-1876-cql-db-node-c0385ef4-2 (34.243.34.213 | 10.4.3.227) (shards: 7)
- longevity-50gb-12h-1876-cql-db-node-c0385ef4-1 (54.247.225.181 | 10.4.1.220) (shards: 5)

OS / Image: `ami-0b71ff818dad1ff1b` (aws: undefined_region)

Test: `longevity-150gb-asymmetric-cluster-12h-test`
Test id: `c0385ef4-0f93-42bd-9ff2-1000310a23d1`
Test name: `scylla-staging/lukasz/longevity-150gb-asymmetric-cluster-12h-test`
[...]

PR #139, opened 2025-05-09
fix(read): fix selecting only configured columns

soyacz:
It was reading with `select *` and causing problems when test added columns (validation errors). Cassandra-stress ignores added columns.

fix by querying specific (configured) columns only. Columns must be quoted as otherwise are lowercased and query fails.

closes: https://github.com/scylladb/cql-stress/issues/138


soyacz:
tested here: https://argus.scylladb.com/tests/scylla-cluster-tests/98523e84-9f4e-4e7e-a25c-7efa3f5d7cdc/nemesis
there's add drop column nemesis and it passed.

muzarski:
Good catch

vponomaryov (approved):

PR #181, opened 2026-03-30
[DO NOT MERGE] custom load balancing policy for strong consistency

Jadw1:
This is a draft PR as there are no intentions to merge it to master (and the commits aren't written in master-grain quality).
Purpose of this patch is to create a temporary custom build of cassandra-stress to do early performance testing of Scylla's strong consistency (the feature is still in experimental phase and it'll be for some time).

The main goal of this changes is to direct request to node which is running Raft leader of tablet group containing request's token.
This is achieved using custom load balancing policy in Rust driver.

With strong consistency, each tablet has its own Raft group with own Raft leader.
So, given a token of the request, the load balancing policy needs to:
- determine to which tablet the token belongs
- what is raft group of the tablet
- which host is the leader of the raft group
- direct the request to this host

Information on which shard the raft group and leader are running can be retrieved by crossing information about  tablet's replicas set (list of `(host, shard)`) and leader's host.

The load balancing policy uses a static `token -> (host, shard)` mapping.
To construct the policy, the `system.tablets` is queried for `last_token -> (raft_group_id, replicas_set)` mapping.
Then, for each raft_group_id `/raft/leader_host` REST endpoint is queried to get the host on which the Raft leader is running and target shard is extracted from `replicas_set`.

The policy is used by default so this build won't work for non-strong-consistent keyspace (it should fail during policy creation).

Fixes https://scylladb.atlassian.net/browse/SCYLLADB-1101

Note that this code was generated by AI!

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
+++ b/src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
@@ -24,9 +24,9 @@ use uuid::Uuid;
 pub struct RaftLeaderPolicy {
     target_keyspace: String,
     target_table: String,
-    /// Mapping from last_token -> leader host_id.
+    /// Mapping from last_token -> (leader host_id, shard).
     /// Populated after session creation via `initialize()`.
-    token_to_leader: OnceLock<BTreeMap<i64, Uuid>>,
+    token_to_leader: OnceLock<BTreeMap<i64, (Uuid, Shard)>>,
     /// Fallback policy for non-target-table queries or when not yet initialized.
     default_policy: Arc<dyn LoadBalancingPolicy>,
 }
@@ -66,14 +66,14 @@ impl RaftLeaderPolicy {
             get_table_id(session, &self.target_keyspace, &self.target_table).await?;
         println!("  Table ID: {}", table_id);
 
-        // Step 2: Get token -> raft_group_id mapping from system.tablets
+        // Step 2: Get token -> (raft_group_id, replicas) mapping from system.tablets
         let tablet_info = get_tablet_info(session, table_id).await?;
         println!("  Found {} tablets", tablet_info.len());
 
         // Step 3: For each unique raft_group_id, get the leader host_id via REST API
         let http_client = reqwest::Client::new();
         let unique_groups: std::collections::HashSet<Uuid> =
-            tablet_info.iter().map(|(_, group_id)| *group_id).collect();
+            tablet_info.iter().map(|t| t.raft_group_id).collect();
         println!(
             "  Querying raft leaders for {} groups...",
             unique_groups.len()
@@ -93,11 +93,27 @@ impl RaftLeaderPolicy {
             group_to_leader.insert(*group_id, leader_host_id);
         }
 
-        // Build the final token -> leader host_id mapping
+        // Build the final token -> (leader host_id, shard) mapping.
+        // For each tablet, find the replica on the leader host to get the shard.
         let mut token_to_leader = BTreeMap::new();
-        for (last_token, group_id) in &tablet_info {
-            if let Some(&leader) = group_to_leader.get(group_id) {
-                token_to_leader.insert(*last_token, leader);
+        for tablet in &tablet_info {
+            if let Some(&leader_host_id) = group_to_leader.get(&tablet.raft_group_id) {
+                let shard = tablet
+                    .replicas
+                    .iter()
+                    .find(|(host_id, _)| *host_id == leader_host_id)
+                    .map(|(_, shard)| *shard as Shard)
+                    .with_context(|| {
+                        format!(
+                            "Leader {} not found in replicas for tablet with last_token {}",
+                            leader_host_id, tablet.last_token
+                        )
+                    })?;
+                println!(
+                    "    Token {} -> Host {} Shard {}",
+                    tablet.last_token, leader_host_id, shard
+                );
+                token_to_leader.insert(tablet.last_token, (leader_host_id, shard));
             }
         }
 
@@ -113,13 +129,13 @@ impl RaftLeaderPolicy {
         Ok(())
     }
 
-    /// Find the leader host_id for a given token.
+    /// Find the leader host_id and shard for a given token.
     ///
     /// Uses the BTreeMap to find the tablet whose range contains the token.
     /// The tablet ranges are defined by `last_token` boundaries:
     /// a token T belongs to the tablet with the smallest `last_token >= T`.
     /// If T is beyond the largest `last_token`, it wraps around to the first tablet.
-    fn find_leader_host_id(&self, token_value: i64) -> Option<Uuid> {
+    fn find_leader(&self, token_value: i64) -> Option<(Uuid, Shard)> {
         let map = self.token_to_leader.get()?;
         if map.is_empty() {
             return None;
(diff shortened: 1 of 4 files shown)


Jadw1:
@Lorak-mmk @wprzytula Could you guys review this PR from Rust driver pov?

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
```diff
@@ -0,0 +1,356 @@
...
+///
+/// This is intended for testing ScyllaDB's strong consistency feature,
+/// where each tablet has its own Raft group and requests should be directed
+/// to the Raft leader of the tablet that owns the requested token.
+///
+/// The mapping is static and set once during initialization. It assumes
+/// that tablet topology won't change (no splits/merges/migrations) and
+/// that the Raft leader won't change during the test.
+#[derive(Debug)]
+pub struct RaftLeaderPolicy {
```
wprzytula, line 24: ❓ When and how are these assumptions going to be relaxed?

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
```diff
@@ -0,0 +1,356 @@
...
+
+use anyhow::{Context, Result};
+use scylla::client::session::Session;
+use scylla::cluster::{ClusterState, NodeRef};
+use scylla::policies::load_balancing::{FallbackPlan, LoadBalancingPolicy, RoutingInfo};
+use scylla::routing::Shard;
+use uuid::Uuid;
+
+/// A load balancing policy that routes requests for a specific table
+/// to the Raft leader of the corresponding tablet.
```
wprzytula, line 12: 🤔 Now the LBP is exclusively for a specific table. I suppose it's going to be a generic solution for all strongly consistent tables in the schema, and information about their strong consistency will be read from the `system_schema` contents?

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
```diff
@@ -0,0 +1,356 @@
...
+        println!("  Found {} tablets", tablet_info.len());
+
+        // Step 3: For each unique raft_group_id, get the leader host_id via REST API
+        let http_client = reqwest::Client::new();
+        let unique_groups: std::collections::HashSet<Uuid> =
+            tablet_info.iter().map(|(_, group_id)| *group_id).collect();
+        println!(
+            "  Querying raft leaders for {} groups...",
+            unique_groups.len()
+        );
```
wprzytula, line 80: ⛏️ Even for such drafts, using `tracing::debug` would look nicer.

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
```diff
@@ -0,0 +1,356 @@
...
+            "  Raft leader mapping initialized: {} tablet entries",
+            token_to_leader.len()
+        );
+
+        self.token_to_leader
+            .set(token_to_leader)
+            .map_err(|_| anyhow::anyhow!("RaftLeaderPolicy already initialized"))?;
+
+        Ok(())
+    }
```
wprzytula, line 130: 💭 This indeed assumes static allocation of strongly consistent tables and tablets. This, being a unrealistic assumption in real use cases, makes me unable to tell what exactly I should review.

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
```diff
@@ -0,0 +1,356 @@
...
+        // For the target table, try to route to the raft leader
+        if self.is_target_table(request) {
+            if let Some(token) = request.token {
+                if let Some(leader_id) = self.find_leader_host_id(token.value()) {
+                    if let Some(result) = Self::find_node_by_host_id(leader_id, cluster) {
+                        return Some(result);
+                    }
+                }
+            }
+        }
```
wprzytula: ♻️ Tip: you can now (with recent Rust versions) use logic operators in `if let` instead of nesting them.

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/// Query system.tablets to get (last_token, raft_group_id) pairs for a table.
+async fn get_tablet_info(session: &Session, table_id: Uuid) -> Result<Vec<(i64, Uuid)>> {
+    // Format the UUID directly in the query since system virtual tables
+    // may not support bind markers well.
+    let query = format!(
+        "SELECT last_token, raft_group_id FROM system.tablets WHERE table_id = {}",
+        table_id
+    );
```
wprzytula: ❓ Virtual tables? I don't think we have such in ScyllaDB.
Also, bind markers in queries to system tables work correctly IIRC.

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
Jadw1: This build is only to allow performance testing of the early version of strong consistency. Currently scylla code doesn't support tablet migration, Raft leader change, splits, merges of strong consistent tables.

This is because the feature is still in experimental phase. In final version you would need to add support to the driver to be aware of tablet Raft group leaders (I guess sth similar to tablet awareness).

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
Jadw1: You will need to implement the generic solution in the driver itself :satisfied:

Information about raft group ids and `token -> raft_group_id` mapping for some particular table can be learned from `system.tablets` table.
But AFAIK currently there is no information which node is a leader for a specific Raft group in any system table. That's why I'm using `/raft/leader_host` REST endpoint to get this information.

Also note that after https://github.com/scylladb/scylladb/pull/27517,
directing request to appropriate Raft leader isn't necessary but this introduces some overhead and we want to avoid it for performance testing.

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
Jadw1: The code was generated by AI and I probably should look closer at things like this. I'll change it.

src/bin/cql-stress-cassandra-stress/raft_leader_policy.rs
Jadw1: We're going to use this build only to do early performance testing of strong consistency (I guess until drivers will add support for strong consistency).
That's why there is only static allocation with restrictions. 
We just need a primitive tool to do strong-consistency-aware cassandra-stress.

I improved the cover letter to better describe purpose of this PR.

wprzytula: In this case, what am I supposed to review?

Jadw1: Is the load balancing correctly implemented.

I guess there is not much to review here, the main thing is logic of the load balancing which is more on our (core devs) side but I'm not familiar with Rust driver that much, so I added you just in case.

Jadw1:
I've tested this build on SCT already and everything is working correctly. Closing this PR to not clutter the space.

soyacz:
> I've tested this build on SCT already and everything is working correctly. Closing this PR to not clutter the space.

what are future plans for it?

piodul:
> > I've tested this build on SCT already and everything is working correctly. Closing this PR to not clutter the space.
> 
> what are future plans for it?

@soyacz Eventually abandon it after we get proper support for leader awareness in relevant drivers. For now, we use this custom version of cql-stress in strong consistency tests.

PR #190, opened 2026-06-16
chore: upgrade scylladb driver to 1.7.0

CodeLieutenant:
Closes https://scylladb.atlassian.net/browse/QATOOLS-240

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/main.rs
+++ b/src/bin/cql-stress-cassandra-stress/main.rs
@@ -164,6 +164,17 @@ async fn prepare_run(
         builder = builder.tls_context(Some(ssl_ctx));
     }
 
+    // Set the preferred datacenter/rack on the SessionBuilder (the recommended way).
+    // The token-aware policy below leaves its own preference unset, so it falls back
+    // to the session-level preference configured here.
+    if let Some(dc) = settings.node.datacenter.clone() {
+        builder = if let Some(rack) = settings.node.rack.clone() {
+            builder.prefer_datacenter_and_rack(dc, rack)
+        } else {
+            builder.prefer_datacenter(dc)
+        };
+    }
+
     let default_exec_profile = ExecutionProfile::builder()
         .load_balancing_policy(settings.node.load_balancing_policy())
         .build();
--- a/src/bin/cql-stress-cassandra-stress/settings/option/node.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/option/node.rs
@@ -16,6 +16,7 @@ pub struct NodeOption {
     pub nodes: Vec<String>,
     pub whitelist: bool,
     pub datacenter: Option<String>,
+    pub rack: Option<String>,
 }
 
 impl NodeOption {
@@ -42,13 +43,26 @@ impl NodeOption {
         println!("  Nodes: {:?}", self.nodes);
         println!("  Is White List: {}", self.whitelist);
         println!("  Datacenter: {:?}", self.datacenter);
+        println!("  Rack: {:?}", self.rack);
     }
 
     fn from_handles(handles: NodeParamHandles) -> Result<NodeOption> {
         let datacenter = handles.datacenter.get();
+        let rack = handles.rack.get();
         let whitelist = handles.whitelist.get().is_some();
         let file = handles.file.get();
         let nodes = handles.nodes.get();
+
+        if datacenter.as_deref() == Some("") {
+            anyhow::bail!("`datacenter` must not be empty");
+        }
+        if rack.as_deref() == Some("") {
+            anyhow::bail!("`rack` must not be empty");
+        }
+        anyhow::ensure!(
+            rack.is_none() || datacenter.is_some(),
+            "`rack` requires `datacenter` to also be set"
+        );
 
         let nodes = match nodes {
             Some(nodes) => nodes,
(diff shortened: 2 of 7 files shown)


Lorak-mmk:
I assume cql-stress has some way of setting preferred DC / rack, right?
In 1.7.0 we introduced an options to set those on SessionBuilder, rather than on DefaultPolicy.
Ofc setting on DefaultPolicy still works, but we do recommend SessionBuilder: in future versions we will be able to use DC/Rack set that way to e.g. prefer using them also for control connection.


See `SessionBuilder::prefer_datacenter_and_rack` and `SessionBuilder::prefer_datacenter`

CodeLieutenant:
> I assume cql-stress has some way of setting preferred DC / rack, right? In 1.7.0 we introduced an options to set those on SessionBuilder, rather than on DefaultPolicy. Ofc setting on DefaultPolicy still works, but we do recommend SessionBuilder: in future versions we will be able to use DC/Rack set that way to e.g. prefer using them also for control connection.
> 
> See `SessionBuilder::prefer_datacenter_and_rack` and `SessionBuilder::prefer_datacenter`

@Lorak-mmk Ok good, I'll add this to this PR, will ring you when it's done

CodeLieutenant:
@Lorak-mmk can you review?

src/bin/cql-stress-cassandra-stress/settings/option/node.rs
```diff
@@ -97,7 +109,14 @@ fn prepare_parser() -> (ParamsParser, NodeParamHandles) {
...
+        "Preferred datacenter, set on the SessionBuilder (driver 1.7.0+)",
+        false,
+    );
+    let rack = parser.simple_param(
+        "rack=",
+        None,
+        "Preferred rack within the preferred datacenter, set on the SessionBuilder \
+        (driver 1.7.0+); requires `datacenter`",
         false,
     );
```
wprzytula, line 126: 🤔 I don't get what the mention `(driver 1.7.0+)` is for. It adds no value IMO.

src/bin/cql-stress-cassandra-stress/main.rs
```diff
@@ -164,6 +164,17 @@ async fn prepare_run(
...
+    // Set the preferred datacenter/rack on the SessionBuilder (recommended since
+    // driver 1.7.0). The token-aware policy below leaves its own preference unset,
+    // so it falls back to the session-level preference configured here.
+    match (&settings.node.datacenter, &settings.node.rack) {
+        (Some(dc), Some(rack)) => {
+            builder = builder.prefer_datacenter_and_rack(dc.clone(), rack.clone())
+        }
+        (Some(dc), None) => builder = builder.prefer_datacenter(dc.clone()),
+        _ => {}
+    }
```
wprzytula: 🔧 It seems these clones are redundant.

src/bin/cql-stress-scylla-bench/args.rs
```diff
@@ -162,6 +166,12 @@ where
         "set the driver host selection policy \
         (round-robin,token-aware,dc-aware:name-of-local-dc),default 'token-aware'",
     );
+    let rack = flag.string_var(
+        "rack",
+        "",
+        "preferred rack within the preferred datacenter, applied on the SessionBuilder \
+        (driver 1.7.0+); requires a 'dc-aware:<dc>' host-selection-policy",
+    );
```
wprzytula, line 174: 🔧 Ditto about `(driver 1.7.0+)` notice.

src/bin/cql-stress-scylla-bench/main.rs
```diff
@@ -140,6 +140,17 @@ async fn prepare(args: Arc<ScyllaBenchArgs>, stats: Arc<ShardedStats>) -> Result
...
+    // Set the preferred datacenter/rack on the SessionBuilder (recommended since
+    // driver 1.7.0). The host selection policy leaves its own preference unset, so it
+    // falls back to the session-level preference configured here.
+    match (&args.datacenter, &args.rack) {
+        (Some(dc), Some(rack)) => {
+            builder = builder.prefer_datacenter_and_rack(dc.clone(), rack.clone())
+        }
+        (Some(dc), None) => builder = builder.prefer_datacenter(dc.clone()),
+        _ => {}
+    }
```
wprzytula: ditto about clones

wprzytula (changes requested):

src/bin/cql-stress-cassandra-stress/main.rs
CodeLieutenant: Actually we cannot do it without clone, since storing it in side the Settings as Option<String> we cannot move it out of settings, we have to clone it everytime, I tried every possible way to remove the clone, but not possible

wprzytula: Ah, I see. Sorry for misguiding you.

src/bin/cql-stress-scylla-bench/main.rs
```diff
@@ -140,6 +140,17 @@ async fn prepare(args: Arc<ScyllaBenchArgs>, stats: Arc<ShardedStats>) -> Result
...
+    // Set the preferred datacenter/rack on the SessionBuilder (the recommended way).
+    // The host selection policy leaves its own preference unset, so it falls back to
+    // the session-level preference configured here.
+    if let Some(dc) = &args.datacenter {
+        let dc = dc.clone();
+        builder = match &args.rack {
+            Some(rack) => builder.prefer_datacenter_and_rack(dc, rack.clone()),
+            None => builder.prefer_datacenter(dc),
+        };
+    }
```
wprzytula: ♻️ Actually the previous approach (using one `match` on `(dc, rack)` pair) was cleaner.

wprzytula (approved):

PR #193, opened 2026-08-13
Leader-aware routing for strongly consistent (Raft-per-tablet) tables

aleksbykov:
# Leader-aware routing for strongly consistent (Raft-per-tablet) tables

Lets `cql-stress-cassandra-stress` benchmark ScyllaDB **strongly consistent** tables — and,
just as importantly, refuse to produce numbers when it did not.

Fixes: SCYLLADB-4583

## Why

Each tablet of a strongly consistent table has a Raft group with a leader that coordinates
writes and most reads. A driver that spreads requests across replicas sends most of them to
followers, which bounce them to the leader — an extra hop that destroys the measurement.

scylla-rust-driver [#1788](https://github.com/scylladb/scylla-rust-driver/pull/1788) does the
routing natively: the server puts the Raft leader first in the `TABLETS_ROUTING_V2` replica
list and `DefaultPolicy` routes there off the driver's own tablet cache. Negotiation is
automatic — **no routing code here**, and less code than the earlier `raft_leader_policy.rs`
workaround (no REST leader lookup, no frozen `token → leader` snapshot, no `reqwest`). What is
left is the part the driver does not do: creating the keyspace, and refusing to run when the
measurement would not mean what it claims.

## Driver: released 1.9.0, behind an off-by-default feature

[v1.9.0](https://github.com/scylladb/scylla-rust-driver/releases/tag/v1.9.0) (2026-09-15) is
the first release containing #1788/#1887/#1890. `[patch.crates-io]` is **gone** — `scylla` is
an ordinary versioned crate resolved from crates.io, with nothing left to rebase.

That release is also what makes the feature gate @CodeLieutenant asked for expressible. Cargo
validates a dependency's features at *resolution* time whether or not the feature naming them
is enabled, so against 1.8.0 even declaring this **with the feature off** broke the default
build for everyone. It resolves now:

```toml
[features]
default = ["user-profile"]
strong-consistency = ["scylla/unstable-strong-consistency"]   # off by default
```

So **everyday builds link no unstable driver surface at all** — `ConsistencyMode` and
`Keyspace::consistency_mode` are not frozen, and the binary SCT runs day to day should not
depend on them. A build without the feature refuses `-schema replication(consistency=...)` at
parse time rather than running unverified: the DDL is just string building and would have kept
working, but the check that proves the run reached leaders needs the driver API, and emitting
the DDL without the check is exactly the plausible-but-meaningless result set this PR exists to
prevent.

The image is built with `--build-arg CARGO_BUILD_FEATURES=strong-consistency`; CI covers both
configurations and `strong-consistency-tests` consumes its own feature-enabled artifact.
Release artifacts are built **without** the feature. `--cfg scylla_unstable` is still required
alongside the crate feature (verified), so the RUSTFLAGS plumbing stays — inert unless the
feature is on.

## Usage

```bash
cql-stress-cassandra-stress write cl=QUORUM duration=10m \
  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3,consistency=global)' \
  -rate threads=100 -node <ip> -log coordinators=true
```

`consistency` is a cql-stress extension to `replication(...)`: not passed into the CQL
replication map, but lifted out into the top-level keyspace property, emitting
`... AND consistency = 'global';`.

**Backward compatibility is the contract.** Without `consistency` the clause is omitted and the
DDL is byte-for-byte what master emits (pinned by a unit test). A default of `'eventual'` would
fail on every server lacking the experimental feature. `local` is rejected at parse time.

## The startup check

The mode comes from the **driver**, not `system_schema`. The driver reports `Global` only once
it has *both* negotiated `TABLETS_ROUTING_V2` *and* read `consistency = 'global'` — it does not
even select that column otherwise. One value proves both capabilities, plus a third no
server-side query can see: that *this driver build* supports leader-aware routing at all.

That is why querying `system_schema` plus probing the `SUPPORTED` frame — the earlier design —
was not merely more code but strictly weaker: both interrogate the server, and neither notices
a driver without #1788, which reads back identical rows and then spreads every request.

Each of these otherwise yields a full, plausible, meaningless result set:

- `CREATE KEYSPACE IF NOT EXISTS` no-ops over a leftover eventually consistent keyspace;
- a `read`-only run against a keyspace never created as global;
- no `--experimental-features=strongly-consistent-tables`;
- the cluster feature not on yet — it turns on only once **every** node has the flag;
- a server without `TABLETS_ROUTING_V2_EXPERIMENTAL`;
- a driver build without leader-aware routing.

The failure enumerates these and asks the contact nodes whether they advertise the extension,
so the report names the missing half. Failure-path only — a healthy run opens no extra
connection. **Hard failure only when the user asked for `consistency=global`**; an
already-global keyspace has its mode reported and `cl=` validated, never failed.

`TABLETS_ROUTING_V2` and strongly consistent tables are **not the same switch**. Verified
against real containers: `scylladb/scylla:2026.2.4` accepts and stores `consistency = 'global'`
while advertising only `TABLETS_ROUTING_V1` — the keyspace reads back as `global` from `cqlsh`
*and* `system_schema.scylla_keyspaces`, the driver correctly reports `Eventual`, and the run
aborts. That trap is why this exists.

## Consistency levels

Far fewer are accepted than on an ordinary keyspace, and the server rejects **per request**, so
the wrong `cl=` yields a run where every operation fails — numbers that read like a catastrophic
cluster problem and are really a CLI mistake. Measured on 2026.4.0~dev:

| | accepted |
|---|---|
| writes | `QUORUM`, `LOCAL_QUORUM` |
| reads | `QUORUM`, `LOCAL_QUORUM`, `ONE`, `LOCAL_ONE` |

`ONE`/`LOCAL_ONE` are legal for reads but **turn leader-aware routing off** — and `local_one` is
the default `cl`, which is what makes this worth enforcing. So: illegal for both → startup
failure; `ONE`/`LOCAL_ONE` on a writing command → startup failure; on `read`/`counterread` →
warning. User profiles carry per-statement levels, so those are validated individually where
they are prepared.

## Preferred datacenter narrows leader routing

A tablet's leader can be in any datacenter, and `cql-stress` never calls
`.permit_dc_failover(true)` — so with `-node datacenter=…` the driver will not contact a remote
leader at all. Those requests go to a local replica and are forwarded: the exact hop this
removes. Roughly **1/N** of the load keeps the benefit with leaders spread over N datacenters,
and nothing in the numbers shows it — the mode still reads `Global`. Hence a startup warning,
fired only when the keyspace is `Global`, `datacenter=` is set, and the cluster spans more than
one. A preferred **rack** never triggers it: the leader outranks rack.

Not a driver bug — with failover disabled the user has said "never contact another datacenter",
and the policy honours that over its own leader preference.

## Verification

| ScyllaDB | `consistency=global` |
|---|---|
| 2025.4.10 | ❌ `Unknown property 'consistency'` |
| 2026.1.10 / 2026.2.3 | ❌ `strongly_consistent_tables feature must be enabled` |
| 2026.2.4, flag on | ❌ startup failure: driver reports `Eventual`, probe names the missing extension |
| 2026.4.0~dev, flag on | ✅ `Global`, leader-aware routing enabled |

`scylladb/scylla-nightly` is the **first image advertising `TABLETS_ROUTING_V2_EXPERIMENTAL`**,
so the compose profile defaults to it — on `scylladb/scylla:latest` every test skips forever and
the CI job is green for the wrong reason.

End-to-end against those containers: `write cl=QUORUM` + `consistency=global` → `Global`,
routing enabled, coordinator distribution printed, 0 errors; `cl=local_one` write → startup
failure; `cl=ALL` read → startup failure; `cl=local_one` read → warning, completes; 2026.2.4 →
startup failure naming the missing extension; eventually consistent runs → unchanged. On a real
two-DC cluster the datacenter warning fires and stays silent in the three cases it should.
Earlier: 4 loaders × 2 × 30 min, 500 threads, `cl=QUORUM` — all 8 runs `Global`, 0 errors,
~52.8k op/s per loader.

`cargo fmt`, clippy in six configurations, `--profile dist` with and without the feature, and
the unit tests in both (85 without `strong-consistency`, 97 with, plus 19 scylla-bench and 13
library). `test_can_connect` needs a live ScyllaDB and fails without one, as on master. 5
integration tests, which skip themselves on a server without V2.

## Notes for reviewers

- `--cfg scylla_unstable` lives in `.cargo/config.toml`. `RUSTFLAGS` **replaces** that setting
  rather than adding to it, so `Dockerfile` and the `fetch_extended_version_info` steps repeat
  it. Where it is load-bearing, getting it wrong is a compile error, never a wrong benchmark.
- `ConsistencyMode` is `#[non_exhaustive]`, so the code matches `Global` rather than enumerating
  alternatives: a future variant means "not strongly consistent", not a build break.
- The `SimpleStrategy` default that used to break this against 2026.2+ was fixed separately in
  #198, now merged; this branch is rebased onto it.
- `uuid` became non-optional (`Node::host_id` is a `uuid::Uuid`, and coordinator accounting is
  not gated behind `user-profile`). It unifies with the driver's own, so nothing new enters.
- The first requests to each tablet are not leader-routed — the driver needs to see a V2 payload
  first. Discard a warm-up window on short runs.
- Rebased onto `master` at `e68d588`; the `-errors` option and the `CachedRow` /
  `SubcommandSchedule` refactor landed there and touch the same operation plumbing.
- Commits will be squashed just before merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,8 +101,13 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 # places the tablet's Raft leader first in the V2 replica list and `DefaultPolicy` routes
 # there deterministically. Negotiation is automatic - there is no client-side opt-in.
 #
-# NOTE: this tracks an OPEN, currently conflicting PR. The rev below is pinned deliberately
-# and is no longer the branch head (`dawmd:leader-awareness` has moved on). Revisit when the
-# PR merges and drop this patch in favour of a released `scylla` version.
+# NOTE: this tracks an OPEN PR. The rev below is the current head of `dawmd:leader-awareness`
+# (the branch was rebased onto a newer upstream master, so it is not a descendant of the rev
+# pinned before). Revisit when the PR merges and drop this patch in favour of a released
+# `scylla` version.
+#
+# The patched crate declares version 1.8.0, which the `scylla = "1.7.0"` requirement above
+# accepts; the requirement is deliberately left at 1.7.0 so that dropping this patch resolves
+# to a released version rather than to one that does not exist on crates.io yet.
 [patch.crates-io]
-scylla = { git = "https://github.com/dawmd/scylla-rust-driver", rev = "3377407235bfd452b5a5b3716977b6aa397ed14c" }
+scylla = { git = "https://github.com/dawmd/scylla-rust-driver", rev = "f9b6ec4f291575a6e83c6a04aa80be4f72df654b" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,9 @@ tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 rust-strictmath = "0.1.2"
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_yaml = { version = "0.9", optional = true }
-uuid = { version = "1.23", optional = true }
+# Not optional: coordinator accounting identifies nodes by `Node::host_id`, which is a
+# `uuid::Uuid`, and that is unconditional. Unifies with the driver's own uuid dependency.
+uuid = "1.23"
 
 [build-dependencies]
 reqwest = { version = "0.13", features = ["json", "blocking"] }
@@ -76,7 +78,7 @@ git2 = "0.21"
 
 [features]
 default = ["user-profile"]
-user-profile = ["dep:serde", "dep:serde_yaml", "dep:uuid"]
+user-profile = ["dep:serde", "dep:serde_yaml"]
 timerfd = ["dep:tokio-timerfd"]
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ timerfd = ["dep:tokio-timerfd"]
 ntest = "0.9"
 num-bigint = "0.4"
 bigdecimal = "0.4"
-scylla = { version = "1.6.0", features = [
+scylla = { version = "1.7.0", features = [
     "openssl-010",
     "num-bigint-04",
     "bigdecimal-04",
@@ -93,3 +93,14 @@ scylla = { version = "1.6.0", features = [
 unexpected_cfgs = { level = "warn", check-cfg = [
     'cfg(fetch_extended_version_info)',
 ] }
+
+# Build against scylla-rust-driver PR #1788 (leader-aware TABLETS_ROUTING_V2 routing),
+# which is required to benchmark strongly consistent (Raft-per-tablet) tables: the server
+# places the tablet's Raft leader first in the V2 replica list and `DefaultPolicy` routes
+# there deterministically. Negotiation is automatic - there is no client-side opt-in.
+#
+# NOTE: this tracks an OPEN, currently conflicting PR. The rev below is pinned deliberately
+# and is no longer the branch head (`dawmd:leader-awareness` has moved on). Revisit when the
+# PR merges and drop this patch in favour of a released `scylla` version.
+[patch.crates-io]
+scylla = { git = "https://github.com/dawmd/scylla-rust-driver", rev = "3377407235bfd452b5a5b3716977b6aa397ed14c" }
--- a/README.md
+++ b/README.md
@@ -59,6 +59,57 @@ by providing it to `ops()` parameter along with other operations defined by the 
 
 To enable the `user` mode, the tool needs to be compiled with `user-profile` feature. This feature is enabled by default.
 
+#### Strongly consistent keyspaces
+
+ScyllaDB supports strongly consistent (Raft-per-tablet) tables, where each tablet has a
+Raft group with a distinguished leader that coordinates writes and most reads. To benchmark
+these meaningfully the client must route requests to the leader rather than spreading them
+across replicas — otherwise most requests land on a follower and are bounced to the leader,
+adding a hop that destroys the measurement.
+
+`cql-stress` can create such a keyspace itself. `consistency` is a **cql-stress extension**
+to the `replication(...)` sub-parameters: it is not passed through into the CQL replication
+map, but lifted out into the top-level `consistency` keyspace property.
+
+```
+cql-stress-cassandra-stress write cl=QUORUM duration=10m \
+  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3,consistency=global)' \
+  -rate threads=100 -node 127.0.0.1
+```
+
+emits
+
+```sql
+CREATE KEYSPACE IF NOT EXISTS "keyspace1"
+  WITH REPLICATION = {'class': 'NetworkTopologyStrategy', 'replication_factor': '3'}
+  AND consistency = 'global';
+```
+
+Accepted values are `global` and `eventual`. When `consistency` is not given the clause is
+omitted from the DDL entirely, so existing invocations are unaffected.
+
+Requirements and caveats:
+
+- The server must run with `--experimental-features=strongly-consistent-tables`. Without it
+  the `consistency` option is rejected outright, and the driver's leader-aware
+  `TABLETS_ROUTING_V2` extension is not advertised at all.
+- The keyspace must be tablet-based. `NetworkTopologyStrategy` enables tablets by default on
+  recent ScyllaDB; `SimpleStrategy` keyspaces may not get tablets, and non-tablet keyspaces
+  reject the `consistency` option.
+- Leader routing is **disabled at `cl=one` and `cl=local_one`**, which keep normal spread
+  routing. Note that `local_one` is the default — pass `cl=QUORUM` explicitly.
+- `CREATE KEYSPACE IF NOT EXISTS` will not upgrade a pre-existing eventually consistent
+  keyspace. Drop it between consistency-mode changes.
+- Only `write`/`counterwrite` create the keyspace; a `read`-only run requires it to already
+  exist.
+
+When `consistency=global` is requested, `cql-stress` reads the keyspace's consistency mode
+back from the driver's cluster metadata at startup and **fails immediately** if it is not
+`Global`, rather than producing plausible but meaningless numbers.
+
+To verify that requests really are concentrated on tablet leaders, enable the per-coordinator
+histogram with `-log coordinators=true`; see [Verifying leader routing](#verifying-leader-routing).
+
 ## Development
 
 ### Prerequisites
--- a/README.md
+++ b/README.md
@@ -107,8 +107,33 @@ When `consistency=global` is requested, `cql-stress` reads the keyspace's consis
 back from the driver's cluster metadata at startup and **fails immediately** if it is not
 `Global`, rather than producing plausible but meaningless numbers.
 
-To verify that requests really are concentrated on tablet leaders, enable the per-coordinator
-histogram with `-log coordinators=true`; see [Verifying leader routing](#verifying-leader-routing).
+#### Verifying leader routing
+
+Pass `-log coordinators=true` to tally operations per coordinator host. The distribution is
+printed in the run summary:
+
+```
+Operations per coordinator:
+  172.17.0.3:9042                   412031 ( 68.4%)
+  172.17.0.2:9042                   109882 ( 18.2%)
+  172.17.0.4:9042                    80512 ( 13.4%)
+```
+
+Interpreting it:
+
+- On a **strongly consistent** keyspace at `cl=quorum`, the distribution must follow the
+  leader distribution across tablets. It will not be uniform.
+- On an **eventually consistent** keyspace — the control — the same workload spreads across
+  all replicas.
+- On a strongly consistent keyspace at `cl=local_one`, the distribution also spreads, because
+  the driver disables leader routing at that level.
+
+Run the control comparison; a single distribution in isolation does not prove much. Accounting
+is off by default and the per-coordinator map is not touched at all when disabled.
+
+Note that the driver only learns a tablet's leader ordering after it has seen a
+`TABLETS_ROUTING_V2` payload for it, so the first requests to each tablet are not leader-routed.
+Discard a warm-up window before drawing conclusions from a short run.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -91,8 +91,14 @@ omitted from the DDL entirely, so existing invocations are unaffected.
 Requirements and caveats:
 
 - The server must run with `--experimental-features=strongly-consistent-tables`. Without it
-  the `consistency` option is rejected outright, and the driver's leader-aware
-  `TABLETS_ROUTING_V2` extension is not advertised at all.
+  the `consistency` option is rejected outright.
+- **The server must also advertise the `TABLETS_ROUTING_V2_EXPERIMENTAL` protocol
+  extension.** These are two separate capabilities and a released ScyllaDB can have the
+  first without the second — ScyllaDB 2026.2.3 accepts `consistency = 'global'` but only
+  advertises `TABLETS_ROUTING_V1`. Without V2 the driver never receives a leader-ordered
+  replica list, so no leader routing happens, and the same gate makes it report every
+  keyspace as eventually consistent. The startup check below turns this into a hard failure
+  rather than a silently meaningless run. A server build with the V2 extension is required.
 - The keyspace must be tablet-based. `NetworkTopologyStrategy` enables tablets by default on
   recent ScyllaDB; `SimpleStrategy` keyspaces may not get tablets, and non-tablet keyspaces
   reject the `consistency` option.
--- a/README.md
+++ b/README.md
@@ -94,10 +94,10 @@ Requirements and caveats:
   the `consistency` option is rejected outright.
 - **The server must also advertise the `TABLETS_ROUTING_V2_EXPERIMENTAL` protocol
   extension.** These are two separate capabilities and a released ScyllaDB can have the
-  first without the second — ScyllaDB 2026.2.3 accepts `consistency = 'global'` but only
+  first without the second — ScyllaDB 2026.2.x accepts `consistency = 'global'` but only
   advertises `TABLETS_ROUTING_V1`. Without V2 the driver never receives a leader-ordered
-  replica list, so no leader routing happens, and the same gate makes it report every
-  keyspace as eventually consistent. The startup check below turns this into a hard failure
+  replica list, so no leader routing happens and every request is bounced to the leader by
+  whichever replica received it. The startup check below turns this into a hard failure
   rather than a silently meaningless run. A server build with the V2 extension is required.
 - The keyspace must be tablet-based. `NetworkTopologyStrategy` enables tablets by default on
   recent ScyllaDB; `SimpleStrategy` keyspaces may not get tablets, and non-tablet keyspaces
@@ -109,9 +109,20 @@ Requirements and caveats:
 - Only `write`/`counterwrite` create the keyspace; a `read`-only run requires it to already
   exist.
 
-When `consistency=global` is requested, `cql-stress` reads the keyspace's consistency mode
-back from the driver's cluster metadata at startup and **fails immediately** if it is not
-`Global`, rather than producing plausible but meaningless numbers.
+When `consistency=global` is requested, `cql-stress` checks both capabilities at startup and
+**fails immediately** if either is missing, rather than producing plausible but meaningless
+numbers:
+
+- the keyspace's consistency mode is read back from `system_schema.scylla_keyspaces`, and
+  must be `global`;
+- the node is asked for its protocol extensions with a bare `OPTIONS` request, and must
+  advertise `TABLETS_ROUTING_V2_EXPERIMENTAL`.
+
+The run prints `Leader-aware routing: enabled` once both hold. For a keyspace that is already
+strongly consistent without `consistency=global` having been passed, the same findings are
+reported as warnings and the run proceeds. The extension probe opens its own plaintext
+connection, so it is skipped with a warning on a TLS run; verify the routing from the
+coordinator distribution below in that case.
 
 #### Verifying leader routing
 
--- a/src/bin/cql-stress-cassandra-stress/operation/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mod.rs
@@ -24,6 +24,7 @@ pub use mixed::MixedOperationFactory;
 pub use row_generator::RowGeneratorFactory;
 use scylla::response::query_result::QueryResult;
 use scylla::value::{CqlValue, Row};
+use uuid::Uuid;
 
 #[cfg(feature = "user-profile")]
 pub use user::UserOperationFactory;
@@ -37,6 +38,39 @@ use self::row_generator::RowGenerator;
 
 const DEFAULT_TABLE_NAME: &str = "standard1";
 const DEFAULT_COUNTER_TABLE_NAME: &str = "counter1";
+
+/// The outcome of a single successfully executed database operation.
+///
+/// Carries the request's coordinator alongside the control flow so that
+/// [`crate::stats::Stats`] can tally operations per host. That tally is the acceptance
+/// evidence for leader-aware routing: on a strongly consistent keyspace the coordinator
+/// distribution must follow the tablet leaders rather than spread across all replicas.
+///
+/// The coordinator is a `Copy` [`Uuid`] read straight off the [`QueryResult`], so carrying
+/// it costs no allocation - which matters, since this is on the per-operation hot path
+/// (see the constraint documented at `src/configuration.rs`).
+pub struct OperationOutcome {
+    pub control_flow: ControlFlow<()>,
+    /// Host ID of the node that coordinated the request.
+    ///
+    /// `None` for operations that complete without a server round-trip.
+    pub coordinator: Option<Uuid>,
+}
+
+impl OperationOutcome {
+    /// A completed operation that should be followed by another one.
+    fn proceed(coordinator: Option<Uuid>) -> Self {
+        Self {
+            control_flow: ControlFlow::Continue(()),
+            coordinator,
+        }
+    }
+}
+
+/// Host ID of the node that coordinated the request, for per-coordinator accounting.
+fn coordinator_of(query_result: &QueryResult) -> Option<Uuid> {
+    Some(query_result.request_coordinator().node().host_id)
+}
 
 /// A specific CassandraStress operation.
 ///
@@ -60,7 +94,7 @@ const DEFAULT_COUNTER_TABLE_NAME: &str = "counter1";
 pub trait CassandraStressOperation: Sync + Send {
     type Factory: CassandraStressOperationFactory<Operation = Self>;
 
-    fn execute(&self, row: &[CqlValue]) -> impl Future<Output = Result<ControlFlow<()>>> + Send;
+    fn execute(&self, row: &[CqlValue]) -> impl Future<Output = Result<OperationOutcome>> + Send;
     fn generate_row(&self, row_generator: &mut RowGenerator) -> Vec<CqlValue>;
     fn operation_tag(&self) -> &str;
 }
@@ -117,7 +151,7 @@ impl<O: CassandraStressOperation> GenericCassandraStressOperation<O> {
             self.cached_row = None;
         }
 
-        op_result
+        op_result.map(|outcome| outcome.control_flow)
     }
 }
 
--- a/src/bin/cql-stress-cassandra-stress/settings/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/mod.rs
@@ -146,6 +146,11 @@ impl CassandraStressSettings {
             "Requested consistency=global, but keyspace '{keyspace}' reports {mode:?}. \
              This run would not measure strong consistency. Check that:\n\
              - the server runs with --experimental-features=strongly-consistent-tables;\n\
+             - the server advertises the TABLETS_ROUTING_V2_EXPERIMENTAL protocol \
+             extension. This is a separate capability, and a server can accept \
+             consistency='global' without having it (e.g. ScyllaDB 2026.2.3, which only \
+             advertises TABLETS_ROUTING_V1). Without it the driver receives no \
+             leader-ordered replica list and reports every keyspace as eventual;\n\
              - keyspace '{keyspace}' does not already exist as an eventually consistent \
              keyspace (CREATE KEYSPACE IF NOT EXISTS will not upgrade it - drop it first);\n\
              - the keyspace is tablet-based (non-tablet keyspaces reject the consistency \
--- a/src/bin/cql-stress-cassandra-stress/settings/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/mod.rs
@@ -137,6 +137,29 @@ impl CassandraStressSettings {
             None => println!("Keyspace '{keyspace}' consistency mode: unknown (keyspace not found in cluster metadata)"),
         }
 
+        // Leader routing is gated on the request's consistency level: the driver keeps
+        // normal spread routing at ONE/LOCAL_ONE. See `DefaultPolicy::should_route_to_leader`.
+        // `local_one` is the default `cl`, so this is a real drift hazard.
+        //
+        // Keyed on the mode the keyspace actually has, not on what was requested: a
+        // pre-provisioned strongly consistent keyspace is leader-routed whether or not
+        // `consistency=global` was passed, since `CREATE KEYSPACE IF NOT EXISTS` no-ops
+        // over it and the mode is a property of the keyspace, not of the CLI flag.
+        if mode == Some(ConsistencyMode::Global) {
+            let cl = self.command_params.common.consistency_level;
+            if matches!(cl, Consistency::One | Consistency::LocalOne) {
+                println!();
+                println!(
+                    "WARNING: keyspace '{keyspace}' is strongly consistent, but cl={cl} \
+                     disables leader-aware routing - the driver keeps normal spread routing \
+                     at ONE and LOCAL_ONE (see DefaultPolicy::should_route_to_leader). \
+                     Requests will be spread over replicas and bounced to the leader. \
+                     Use cl=QUORUM to measure strong consistency."
+                );
+                println!();
+            }
+        }
+
         if !self.schema.wants_strong_consistency() {
             return Ok(());
         }
@@ -158,22 +181,6 @@ impl CassandraStressSettings {
              DDL used: {ddl}",
             ddl = self.schema.construct_keyspace_creation_query(),
         );
-
-        // Leader routing is gated on the request's consistency level: the driver keeps
-        // normal spread routing at ONE/LOCAL_ONE. See `DefaultPolicy::should_route_to_leader`.
-        // `local_one` is the default `cl`, so this is a real drift hazard.
-        let cl = self.command_params.common.consistency_level;
-        if matches!(cl, Consistency::One | Consistency::LocalOne) {
-            println!();
-            println!(
-                "WARNING: keyspace '{keyspace}' is strongly consistent, but cl={cl} disables \
-                 leader-aware routing - the driver keeps normal spread routing at ONE and \
-                 LOCAL_ONE (see DefaultPolicy::should_route_to_leader). Requests will be \
-                 spread over replicas and bounced to the leader. Use cl=QUORUM to measure \
-                 strong consistency."
-            );
-            println!();
-        }
 
         Ok(())
     }
--- a/src/bin/cql-stress-cassandra-stress/settings/mod.rs
+++ b/src/bin/cql-stress-cassandra-stress/settings/mod.rs
@@ -4,6 +4,7 @@ use std::iter::Iterator;
 mod command;
 mod option;
 mod param;
+mod protocol_extensions;
 use anyhow::Context;
 use anyhow::Result;
 
@@ -20,7 +21,7 @@ pub use option::LogOption;
 pub use option::ThreadsInfo;
 use regex::Regex;
 use scylla::client::session::Session;
-use scylla::cluster::metadata::ConsistencyMode;
+use scylla::errors::{DbError, ExecutionError, RequestAttemptError};
 use scylla::statement::Consistency;
 
 use crate::settings::command::print_help;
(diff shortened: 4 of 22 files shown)


aleksbykov:
@Jadw1 @dawmd @gusev-p @bitpathfinder can you review and approve that cql-stress correctly switched to new driver from PR:  https://github.com/scylladb/scylla-rust-driver/pull/1788  and use it functionality?

piodul:
> @Jadw1 @dawmd @gusev-p @bitpathfinder can you review and approve that cql-stress correctly switched to new driver from PR: [scylladb/scylla-rust-driver#1788](https://github.com/scylladb/scylla-rust-driver/pull/1788) and use it functionality?

To me, it looks like the first commit is enough to make cql-stress use Dawid's leader-aware driver. It should just work if you send requests to a strongly consistent keyspace with it. That commit looks OK.

If you want to use cql-stress to create a strongly consistent keyspace for you, then the second commit is needed and it looks OK as well.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
```diff
@@ -38,6 +39,39 @@ use self::row_generator::RowGenerator;
...
+/// distribution must follow the tablet leaders rather than spread across all replicas.
+///
+/// The coordinator is a `Copy` [`Uuid`] read straight off the [`QueryResult`], so carrying
+/// it costs no allocation - which matters, since this is on the per-operation hot path
+/// (see the constraint documented at `src/configuration.rs`).
+pub struct OperationOutcome {
+    pub control_flow: ControlFlow<()>,
+    /// Host ID of the node that coordinated the request.
+    ///
+    /// `None` for operations that complete without a server round-trip.
```
Jadw1: > `None` for operations that complete without a server round-trip.

but `coordinator_of()` (used to fill in this field) never return None

Jadw1 (commented):
Looks like the driver is correctly used in terms of the strong consistency feature.
I don't know cql-stress well enough to fully asses this PR, I only found one nit-pick.

src/bin/cql-stress-cassandra-stress/operation/mod.rs
aleksbykov: Right - the `None` case was unreachable: `coordinator_of` reads `request_coordinator().node().host_id` off a `QueryResult`, and every `OperationOutcome` is built from a completed round-trip, so there is no call site that can produce it. The doc comment described a case that does not exist.

Fixed in `1dc9040`: `coordinator` is now a plain `Uuid`, and the paired `if let (Some(coordinators), Some(host_id))` in `Stats::account_operation` collapses to a single `Some` on the `-log coordinators=true` toggle. Thanks for catching it.

aleksbykov:
@Jadw1 thanks - driver-side correctness is the half I most wanted your eyes on. The nit is addressed in `1dc9040`.

One thing worth your input while you're here: #1788 keeps `ConsistencyMode` and `Keyspace::consistency_mode` `pub(crate)` (still true at `cbe2929`), and `internal_testing` does not expose the mode either, so there is no public way to ask the driver whether V2 was negotiated or what mode a keyspace is in. That is why this PR reads `system_schema.scylla_keyspaces.consistency` itself and probes the extension with a bare `OPTIONS` request. If #1788 ends up exposing an accessor (`is_strongly_consistent` is referenced in the docs but does not exist), both checks collapse into it - the driver only reads that column when V2 was negotiated, so a `Global` mode would prove both halves at once. Is exposing it on the table?

aleksbykov:
@piodul agreed on both, and that is exactly the split: `deps:` alone gets leader-aware routing, `schema:` is only needed so cql-stress can create the keyspace itself.

The rest of the PR exists for a third reason - proving the run measured what it claims. Two independent server capabilities are involved, and a released ScyllaDB can have one without the other: **2026.2.x accepts `consistency = 'global'` and records it in `system_schema.scylla_keyspaces`, but advertises only `TABLETS_ROUTING_V1`** (re-verified on `scylladb/scylla:2026.2.4`). Such a run looks strongly consistent from every angle and is not leader-routed at all - and `CREATE KEYSPACE IF NOT EXISTS` silently no-opping over a leftover eventually consistent keyspace, or a `read`-only run against one, fail the same way: a full, plausible, meaningless result set. So the mode is checked against `system_schema` and the extension against the node's `SUPPORTED` frame, both hard failures only when `consistency=global` was actually requested; `-log coordinators=true` then gives the per-coordinator distribution as positive evidence.

Also flagging a pre-existing issue you may want to weigh in on: the default `-schema` is `SimpleStrategy`, which 2026.2+ rejects outright (`SimpleStrategy doesn't support tablet replication`). That already breaks two integration tests and CI against current `scylladb/scylla` on `master`. Fixing the default feels like a separate PR - agree?

fruch:
@aleksbykov now that https://github.com/scylladb/scylla-rust-driver/pull/1788 is merged (probably would need it released as well), I would recommend  working with @CodeLieutenant and @soyacz to get the rest of change here reviewed (and not wait for the rust driver release for that)

aleksbykov:
Ok, get it. I will do that

Cargo.toml
```diff
@@ -93,3 +93,14 @@ scylla = { version = "1.6.0", features = [
...
+# Build against scylla-rust-driver PR #1788 (leader-aware TABLETS_ROUTING_V2 routing),
+# which is required to benchmark strongly consistent (Raft-per-tablet) tables: the server
+# places the tablet's Raft leader first in the V2 replica list and `DefaultPolicy` routes
+# there deterministically. Negotiation is automatic - there is no client-side opt-in.
+#
+# NOTE: this tracks an OPEN, currently conflicting PR. The rev below is pinned deliberately
+# and is no longer the branch head (`dawmd:leader-awareness` has moved on). Revisit when the
+# PR merges and drop this patch in favour of a released `scylla` version.
+[patch.crates-io]
+scylla = { git = "https://github.com/dawmd/scylla-rust-driver", rev = "3377407235bfd452b5a5b3716977b6aa397ed14c" }
```
dawmd: Note: Leader-awareness has been merged and it should be released very soon (~1 week from now AFAIK).

README.md
```diff
@@ -59,6 +59,57 @@ by providing it to `ops()` parameter along with other operations defined by the
...
+```
+
+Accepted values are `global` and `eventual`. When `consistency` is not given the clause is
+omitted from the DDL entirely, so existing invocations are unaffected.
+
+Requirements and caveats:
+
+- The server must run with `--experimental-features=strongly-consistent-tables`. Without it
+  the `consistency` option is rejected outright, and the driver's leader-aware
+  `TABLETS_ROUTING_V2` extension is not advertised at all.
```
dawmd: It might be worth noting the discrepancy in the server: TABLETS_ROUTING_V2 is advertised by the server when the experimental feature is enabled, but strongly consistent tables are gated behind a cluster feature (which itself is gated behind the experimental feature).

README.md
```diff
@@ -59,6 +59,57 @@ by providing it to `ops()` parameter along with other operations defined by the
...
+- The keyspace must be tablet-based. `NetworkTopologyStrategy` enables tablets by default on
+  recent ScyllaDB; `SimpleStrategy` keyspaces may not get tablets, and non-tablet keyspaces
+  reject the `consistency` option.
+- Leader routing is **disabled at `cl=one` and `cl=local_one`**, which keep normal spread
+  routing. Note that `local_one` is the default — pass `cl=QUORUM` explicitly.
```
dawmd: This is also only possible for reads. We cannot performs writes with CL other than `LOCAL_QUORUM` or `QUORUM`.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
```diff
@@ -137,6 +137,29 @@ impl CassandraStressSettings {
             None => println!("Keyspace '{keyspace}' consistency mode: unknown (keyspace not found in cluster metadata)"),
         }
 
+        // Leader routing is gated on the request's consistency level: the driver keeps
+        // normal spread routing at ONE/LOCAL_ONE. See `DefaultPolicy::should_route_to_leader`.
```
dawmd: We'll probably need to go through the references -- the driver has changed a lot ever since its first version, and I'm afraid many of them may be out-of-date. Also, I don't know if it makes sense to mention it here -- this project and the driver are completely separate ones, and the implementation can easily change (I'm not sure about the guarantees given to the user without looking up the code). But this is just me thinking loudly -- someone from the driver team would be a more fitting person to provide input here.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
```diff
@@ -4,6 +4,7 @@ use std::iter::Iterator;
 mod command;
 mod option;
 mod param;
+mod protocol_extensions;
```
dawmd, line 7: commit: main: check strong consistency against the server, not the driver

`ConsistencyMode` is now exposed via an unstable feature: https://github.com/scylladb/scylla-rust-driver/commit/c71ee8c2e1e027c9e76f398c0ab488b2f4908130. Note that we dropped `ConsistencyMode::Local` -- it's not implemented yet, the name can change, etc. It'll only come back once it's all in place.

dawmd (commented):
Sorry I didn't take a look earlier, but I've been flooded with CI issues and other high-priority tasks. I don't have much time now either, but I didn't want to leave this PR without even a single comment. I only looked at the following commits:
* `deps: build against scylla-rust-driver PR #1788 (TABLETS_ROUTING_V2)`
* `schema: support keyspace consistency mode in CREATE KEYSPACE`
   * only the docs
* `stats: account operations per coordinator to verify leader routing`
   * only the docs
* `ci: add strongly-consistent-tables test environment`
   * only the docs
* `main: warn on cl=local_one for any strongly consistent keyspace`
* `main: check strong consistency against the server, not the driver`
   * only the commit message

The rest of the PR seems to revolve cql-stress itself -- I don't have much knowledge about it, but I see other people working on strong consistency have reviewed it, so I believe it's good.

As some general thoughts: we re-exposed `ConsistencyMode`, so perhaps at least some part of this work will be simplified thanks to it -- not sure how, though. :)

Cargo.toml
aleksbykov: ok, great

README.md
aleksbykov: Good catch — the README claimed a single gate. Fixed in `3b8de22`, which now spells out both directions with a table, and the discrepancy is also listed as a cause in the startup failure message.

Verified on containers: `scylladb/scylla-nightly` (2026.4.0~dev) advertises `TABLETS_ROUTING_V2_EXPERIMENTAL` only when started with `--experimental-features=strongly-consistent-tables` and not otherwise, so V2 does follow the per-node flag; and 2026.2.4 with the flag accepts `consistency = 'global'` while advertising V1 only. A rolling enable therefore hits the opposite case — a node advertising V2 while the cluster feature is still off — and that is now enumerated too.

README.md
aleksbykov: Right, and it turned out stricter than a caveat could cover, so this became a hard check in `8fce03f` rather than a doc note.

Measured on 2026.4.0~dev: writes take QUORUM/LOCAL_QUORUM only; reads additionally take ONE/LOCAL_ONE; everything else (ALL, TWO, THREE, ANY, EACH_QUORUM) is rejected for both, per request rather than at connect time. So `write cl=ALL` used to start happily and then fail every single operation.

Now: a level illegal for both fails at startup; ONE/LOCAL_ONE fails at startup for a writing command; a read at ONE/LOCAL_ONE keeps the old warning, since it works but is not leader-routed. A user profile cannot be classified from here, so it warns about both halves. The README carries both accepted sets.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
aleksbykov: Agreed, and they are gone — `8fce03f` removes every reference to a driver-private symbol from this repo. The comments now describe observable behaviour instead, which is what cql-stress can actually rely on across driver versions.

The same commit also removes the reason most of them existed: the mode now comes from `Keyspace::consistency_mode` rather than from cql-stress reimplementing the driver's logic against `system_schema`, so there is much less driver behaviour to describe in the first place.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
aleksbykov: Thanks — that is exactly what this needed, and it collapses the whole thing.

`665a727` repins to upstream `main` (past `c71ee8c2`) and turns on `unstable-strong-consistency` plus `--cfg scylla_unstable`; `8fce03f` then deletes the local `ConsistencyMode` mirror, the `system_schema.scylla_keyspaces` query and `is_missing_table_or_column`, and demotes the `OPTIONS` probe to a failure-path diagnostic.

It is also strictly more correct, not just less code. Both old checks interrogated the server, so neither could notice a *driver* without leader-aware routing — it reads back the same rows and the same SUPPORTED frame. `consistency_mode == Global` proves the driver negotiated V2, since it does not select that column otherwise. Confirmed against 2026.2.4: the server stores 'global' and the driver correctly reports `Eventual`.

`Local` is dropped here too; the parse-time rejection of `consistency=local` stays, since the server still does not implement it.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
```diff
@@ -103,6 +108,294 @@ impl CassandraStressSettings {
...
+            .await
+            .context("Failed to refresh cluster metadata")?;
+
+        let keyspace = &self.schema.keyspace;
+        let cluster_state = session.get_cluster_state();
+        // `None` means the keyspace does not exist at all, which is a different thing from
+        // existing as eventually consistent, and the two get different messages below.
+        let mode = cluster_state
+            .get_keyspace(keyspace)
+            .map(|ks| ks.consistency_mode.clone());
```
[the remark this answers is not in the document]
aleksbykov: Confirmed, and it is a real break rather than a theoretical one. Reproduced locally:

```
$ RUSTFLAGS="--cfg fetch_extended_version_info" cargo check --all
error[E0603]: enum `ConsistencyMode` is private
error[E0616]: field `consistency_mode` of struct `Keyspace` is private
```

`ee4a54e` set `--cfg scylla_unstable` in `rust.yml` and in the `Dockerfile` but missed `release.yml`, whose workflow-level `RUSTFLAGS` replaces `.cargo/config.toml`'s `[build] rustflags` wholesale. `release-artifacts` therefore built without the cfg and would have failed *after* the GitHub release was already created and before any asset was uploaded. `docker_build` was unaffected — it goes through the `Dockerfile`, which repeats both cfgs.

Fixed in `98c7429`: the cfg is added to `release.yml`'s `env:` block, `.cargo/config.toml`'s comment now names all three places that set `RUSTFLAGS` rather than two (which is how this one was missed), and a `--profile dist` build with the release workflow's exact `RUSTFLAGS` is added to the PR matrix. That last part is the actual regression guard — the release profile was otherwise never exercised until a tag was pushed.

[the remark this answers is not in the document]
aleksbykov: Both halves are real, and the second one is real in a way the description does not quite capture — thanks for pushing on it.

**The keyspace is a straight bug.** `verify_consistency_mode` read `self.schema.keyspace` unconditionally, but a user run creates and drives the keyspace declared in its yaml: `UserParams::create_schema` executes the profile's `keyspace_definition`, and `UserOperationFactory::new` reads its table metadata from `user_profile.keyspace`. `-schema` is not consulted at all in user mode. So a user profile against a strongly consistent keyspace got no mode report, no `cl=` check and no datacenter warning — it looked up `keyspace1`, found nothing, printed `unknown` and returned `Ok`. And `user` + `-schema replication(...,consistency=global)` hard-failed on a keyspace nothing in the run would ever have created.

Fixed in `dcacd80`: one effective keyspace is resolved in `workload_keyspace()` and used for the mode lookup, the `cl=` check and the datacenter warning. `consistency=` is now rejected at parse time with the `user` command — it rides on the keyspace creation query user mode never executes, so it was inert, and accepting it silently reads like a request that was honoured while the run measures an eventually consistent keyspace. The error says to put `AND consistency = 'global'` in the profile's `keyspace_definition` instead.

**On the consistency levels, one correction:** the predefined `insert` is not at risk. It sets no statement-level consistency, so it inherits the driver's execution-profile default of `LOCAL_QUORUM`, which is legal for strongly consistent writes — as does any yaml query without `consistencyLevel:`.

The actual gap is narrower and worse: **`cl=` is never applied in user mode at all.** It is read only by `operation/{write,read,counter_write}.rs`. So a yaml query declaring `consistencyLevel: ALL` or `ONE` ran completely unvalidated and failed every operation, and the `None` arm of `verify_consistency_level` — the "warn about both halves" branch — warned about a `cl=` that has no effect on the run. It would fire on `user cl=local_one` even when every query in the profile declared `LOCAL_QUORUM`.

Fixed in `081cec6`: the effective level of each prepared statement is validated in `UserOperationFactory::new` once the keyspace is known to be `Global`. Reads and writes are told apart by the statement's own result metadata rather than by parsing the CQL — a statement that returns columns is a read, except a conditional write, whose result set is the `[applied]` flag and which is therefore treated as a write (the stricter and correct side). Every violation is reported together, naming each query, since a profile can easily have several and fixing them one run at a time is needless. The misleading `None` warning is gone, and `issues_writes()` loses its `Option`.

The rules themselves moved into `classify_consistency`, unit tested over the whole `Consistency` enum — getting them wrong does not fail at connect time, it fails every operation of the run, so it is worth pinning without needing a server built with the experimental feature.

tools/test_cs_strong_consistency.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+        except Exception as e:
+            return f"server does not support the strongly-consistent-tables feature: {e}"
+
+        # n must stay above 1: a single-operation run cannot build its sequence
+        # distribution, and the probe would report every server as unsupported.
+        result = cql_stress.run_raw(
+            stress_args("write", node, probe, cl="QUORUM", consistency="global", n=10),
+            check=False)
+        if "Leader-aware routing: enabled" not in result.stdout:
```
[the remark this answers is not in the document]
aleksbykov: Valid, and it defeats the purpose of the job — a panic, a renamed CLI option, a missing shared library or an unreachable node all came back as "server does not advertise TABLETS_ROUTING_V2_EXPERIMENTAL", the fixture turned that into a skip, and the whole strong-consistency suite went green without running anything.

Fixed in `a701f8e`. The probe now discriminates on the return code and on a **stable diagnostic code** the failure carries, rather than on prose: cql-stress appends `(diagnostic code: STRONG_CONSISTENCY_UNAVAILABLE)` to that one startup failure, and a Rust unit test pins it. Matching a code rather than a sentence means rewording the error breaks a fast unit test instead of silently un-skipping CI months later.

The rules are now:

| probe outcome | result |
|---|---|
| exit 0, success marker present | supported, suite runs |
| nonzero exit carrying the diagnostic code | skip — genuine missing server capability |
| any other nonzero exit | **raise**, with stdout and stderr attached |
| exit 0 without the success marker | **raise** — the probe cannot classify it |

The `CREATE KEYSPACE` half had the same shape: a bare `except Exception` laundered auth failures and timeouts into a skip. It is narrowed to the server's own rejection of the option, and re-raises everything else.

Verified against a stubbed runner covering all four rows above plus the auth-failure case — a broken binary now raises, and an unsupported server still skips.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
```diff
@@ -103,6 +108,294 @@ impl CassandraStressSettings {
...
+        let reported_mode = match &mode {
+            Some(mode) => format!("{mode:?}"),
+            None => String::from("unknown (keyspace not found in cluster metadata)"),
+        };
+        println!("Keyspace '{keyspace}' consistency mode: {reported_mode}");
+
+        // `ConsistencyMode` is `#[non_exhaustive]`: match the one variant that means strong
+        // consistency rather than enumerating the others, so a future variant is treated as
+        // "not strongly consistent" instead of failing to compile.
+        let strongly_consistent = matches!(mode, Some(ConsistencyMode::Global));
```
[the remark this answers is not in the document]
aleksbykov: The mechanism is right, and worth stating precisely: the keyspace consistency column is only selected when the driver's metadata connection reports V2 support, and V2 framing is negotiated per connection. So `Global` proves that *the connection the driver read cluster metadata over* negotiated V2 — not that every node did. A cluster part-way through a rolling enable, far enough for the cluster feature to be on but with nodes still missing the flag, can therefore report `Global` while some workload connections are not leader-routed.

I'm not turning that into a cluster-wide gate, though, for four reasons:

- the probe speaks plaintext CQL, so it is unavailable on exactly the TLS clusters that would need it;
- sweeping every node on startup reverses the deliberate choice to keep the happy path free of extra connections — the `OPTIONS` probe was moved to the failure path for that reason;
- a partially flagged cluster is a legitimate thing to benchmark on purpose; this is a load generator, not an admission controller;
- the driver exposes no public aggregate, and reaching into private symbols is something this PR was already asked to stop doing (see dawmd's earlier comment).

What `024f51a` does instead is narrow the claim to what is actually proven, and name the signal that *does* catch the mixed case. `-log coordinators=true` already prints the per-coordinator distribution, and a mixed cluster shows up there as a spread across replicas instead of a skew toward leaders — so the success line now points at it, the doc comment on `verify_consistency_mode` spells out the limitation, and the README's capability table gains a rolling-enable note.

I'll open a driver issue asking for a public, cluster-aggregate signal for whether leader-aware routing is in effect across the session's connections, and link it here. If that lands, the check tightens for free.

src/bin/cql-stress-cassandra-stress/settings/mod.rs
```diff
@@ -103,6 +108,294 @@ impl CassandraStressSettings {
...
+                 run uses TLS."
+            );
+        }
+
+        match fetch_protocol_features(node).await {
+            Ok(features) if features.tablets_v2_supported => format!(
+                "\nNode '{node}' does advertise TABLETS_ROUTING_V2_EXPERIMENTAL, so this \
+                 server can route to tablet leaders - the keyspace itself is what is not \
+                 strongly consistent."
+            ),
```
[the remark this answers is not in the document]
aleksbykov: Right — the probe took the first `-node` entry while the session is built from all of them, and the positive branch then generalised from that one node to the cluster. That is precisely backwards during a rolling enable, where the first contact point carries the flag, another does not, and the real cause is the cluster feature still being off.

Fixed in `52de177`. The probe now sweeps every configured contact node concurrently and classifies the answers three ways:

- **all** advertise it → the extension is not the missing half, the keyspace is;
- **none** do → no node can hand out a leader-ordered replica list (the 2026.2.x trap);
- **mixed** → the disagreement *is* the diagnosis, and both groups are named: the cluster feature that gates `consistency='global'` stays off until every node carries the flag, so the keyspace cannot be created strongly consistent yet even though some nodes could already route to leaders.

An unreachable node no longer discards what the others answered — it is listed separately. The list is capped at 8 nodes so a large `-node` list does not produce an unreadable error, and the message says how many were left out rather than implying the rest were found to agree.

The wording is split into `summarise_v2_probe` and unit tested over all of those rows, including the mixed one, since the wording is the whole point of a diagnostic and needs no server to check.

aleksbykov:
Pushed `9ed52bb..71784cd`, addressing all five findings from the Qodo review. Individual replies are on each thread; summary here.

| finding | verdict | commit |
|---|---|---|
| 2. Tagged releases fail to build | confirmed, reproduced locally | `98c7429` |
| 1. User profiles bypass leader routing | valid on the keyspace; the CL half is real but different from the description | `dcacd80`, `081cec6` |
| 3. Broken binaries skip the new test suite | confirmed | `a701f8e` |
| 4. Mixed clusters pass routing check | mechanism confirmed; scoped to an honest claim rather than a cluster-wide gate | `024f51a` |
| 5. Startup errors blame the wrong node | confirmed | `52de177` |

Two of these were worth the review on their own:

- **The release break was real.** `RUSTFLAGS` in `release.yml` replaced `.cargo/config.toml`'s rustflags, so every tag build would have failed on `field consistency_mode is private` *after* the GitHub release had been created. `ee4a54e` fixed `rust.yml` and the `Dockerfile` and missed this one. There is now a `--profile dist` build in the PR matrix using the release workflow's exact flags, so the next such slip breaks a PR instead of a tag.
- **User-profile runs were checking the wrong keyspace entirely**, and `cl=` — which user mode never applies to anything — was standing in for per-statement levels it has no relationship to. Both are fixed; the yaml's own `consistencyLevel:` values are now validated per query, with reads and writes told apart from the prepared statement's result metadata.

One correction to finding 1 worth repeating here: the predefined `insert` was never at risk. It carries no statement-level consistency and so inherits the driver's `LOCAL_QUORUM` default, which is legal for strongly consistent writes.

### Driver status

Still no release containing [#1788](https://github.com/scylladb/scylla-rust-driver/pull/1788) — the newest tag remains `v1.8.0`, cut 2026-08-13, before the merge. `[patch.crates-io]` therefore stays, repinned in `9ed52bb` to the current upstream `main` (`ada16b4e`). The moment a release lands the section gets deleted and the requirement moved to it; that cannot be done silently by accident, since `unstable-strong-consistency` does not exist in 1.8.0 and cargo simply fails to resolve. cc @dawmd

### Verification

`cargo fmt --check`, the full clippy matrix (all targets, `--no-default-features`, `user-profile`, `fetch_extended_version_info`), the new `--profile dist` release build, and 129 unit tests (was 122) are green. `test_can_connect` needs a live ScyllaDB and fails without one, as on master.

### Still open, and not part of this PR

- **The `SimpleStrategy` default.** Still unanswered from the 2026-08-31 thread. The default `-schema` is `SimpleStrategy`, which 2026.2+ rejects outright (`SimpleStrategy doesn't support tablet replication`); that already breaks two integration tests and CI against current `scylladb/scylla` **on master**, independently of this branch. It is a backward-compatibility decision about a cassandra-stress-compatible default and deserves its own PR — I'll open a separate issue and link it. @piodul, any objection to that split?
- **A driver-side aggregate for V2 support** (see the reply on finding 4). I'll file it upstream and link it here.

@CodeLieutenant @soyacz — per @fruch's suggestion, the cql-stress half of this still needs a review. The commits that are cql-stress-only rather than driver-facing are `schema:`, `stats:` and `ci:`, plus the five above.

CodeLieutenant:
@aleksbykov Can you please remove the loader-set-* folders? They are from SCT and should not be here. If you need testing, please write integration tests. 

This feature looks like it needs driver support that is unstable, so we should have a separate release for this to avoid messing with the cql-stress that's running in SCT. Putting this behind the feature flags is the way to go. Also, upgrade the driver version to 1.8.0 since that's on master here. Building the driver from source (git commit) must be behind the flag, as introducing unstable features to a stable project is a no go. Tracing bugs for it will be hard, and this commit of the driver must constantly be rebased so that we get new stuff from the releases, as this project will move at the pace of the driver.

Better to have two binaries, one stable and one needed for one purpose then unstable project.

Please tidy up the commits; this is really hard to review.

aleksbykov:
@CodeLieutenant thanks for the review.

**On the `loader-set-*` folders — done, and you're right that they had no business being here.**

They were not test fixtures. They were local SCT run output sitting untracked in my working tree that got swept in by an unscoped `git add -A`. Removed, and the history has since been squashed so they are gone from the branch entirely, not just netted out of the diff — `git log --name-only` over the branch range returns nothing for those paths at the current head (`043ad94`). Sorry for the review noise.

Test coverage is in `tools/test_cs_strong_consistency.py` — five integration tests against a `strongly-consistent-tables` container, wired into a `strong-consistency` CI job that is green on this PR.

**On the feature flag and released 1.8.0 — I agree with the end state, and I checked what it takes to get there. Short version: it is blocked on the driver release, not on willingness.**

I tested both halves rather than guessing.

The patch itself moves out cleanly. `[patch.crates-io]` works from a cargo config file passed with `--config`, so the default build resolves to released `scylla 1.8.0` from crates.io with no git dependency, and only the strong-consistency build passes the extra config. That part is fine.

The feature gate is the problem. Cargo validates dependency features at **resolution** time, whether or not your feature is enabled. With `strong-consistency = ["scylla/unstable-strong-consistency"]` declared but switched **off**, against released 1.8.0:

```
package `cql-stress` depends on `scylla` with feature `unstable-strong-consistency`
but `scylla` does not have that feature.
failed to select a version for `scylla` which could resolve this conflict
```

`unstable-strong-consistency` exists in no released driver, so merely *naming* it in an optional feature breaks the default build for everyone. The one-crate-one-flag design cannot be expressed today.

That leaves three ways forward, and I think the third is clearly right:

1. a second manifest or workspace member — works, but it is scaffolding we delete in a few weeks;
2. merge as-is with the git pin — which is what you are objecting to, and I agree with the objection: you are right that it drags this project along at the driver's pace and makes bugs hard to trace;
3. **hold this PR until a driver release contains `unstable-strong-consistency`.** At that point `[patch.crates-io]` disappears entirely, the dependency becomes an ordinary versioned crate, and `strong-consistency = ["scylla/unstable-strong-consistency"]` resolves correctly with the feature off — exactly the shape you asked for, with no git dependency and nothing to rebase.

#1788 merged upstream on 2026-09-02 and the newest tag is still `v1.8.0`, cut 2026-08-13 — before it. A release was expected "~1 week" as of 2026-09-09. cc @dawmd — any update on timing?

Strong-consistency performance testing is unblocked in the meantime by a custom image built from this branch, so nothing is waiting on the merge. Does holding until the release work for you? If you would rather have the two-binary split landed sooner than that, say so and I will build it.

**On the commits** — the branch is currently 8 commits, one concern each: the driver pin, the CI flags fix, and one per review finding. I will squash it into its final shape once the driver release lands and the `[patch.crates-io]` removal and feature gate are in, since that rework will rewrite the dependency commits anyway — squashing now would only have to be redone then. If the current sequence is hard to follow in the meantime, tell me and I will collapse it sooner.

aleksbykov:
## scylla-rust-driver 1.9.0 is out — everything that was waiting on it is done

[v1.9.0](https://github.com/scylladb/scylla-rust-driver/releases/tag/v1.9.0) was tagged on
2026-09-15 and is the first release containing #1788, #1887 and #1890. Pushed as `b5bd8b6`
and `c5326ca`.

@CodeLieutenant — this is the shape you asked for, and all three of your points are now
actually addressed rather than deferred:

**1. No git dependency.** `[patch.crates-io]` is deleted. `scylla = "1.9.0"` is an ordinary
versioned crate and `Cargo.lock` resolves it from crates.io — no git rev anywhere. Your
concern that "this commit of the driver must constantly be rebased so that we get new stuff
from the releases" no longer applies: there is nothing to rebase.

**2. The unstable API is behind a feature flag, off by default.**

```toml
[features]
default = ["user-profile"]
strong-consistency = ["scylla/unstable-strong-consistency"]   # off by default
```

This is exactly what could not be expressed against 1.8.0, and the reason was resolution
order, not effort: cargo validates a dependency's features when it *resolves* the graph,
whether or not the feature naming them is enabled, so merely declaring that line with the
feature **switched off** failed for everyone with `package cql-stress depends on scylla with
feature unstable-strong-consistency but scylla does not have that feature`. The feature now
exists in a released driver, so it resolves correctly with the feature off. I re-verified
both halves rather than assuming.

The result is what you wanted: **the cql-stress that runs in SCT links no unstable driver
surface at all.** A default build compiles none of the strong-consistency code, and refuses
`-schema replication(consistency=...)` at parse time with a message saying which feature is
missing and how to get a binary that has it. Refusing rather than ignoring is deliberate —
the DDL itself is just string building and would have kept working, but the check that proves
the run reached tablet Raft leaders needs the driver API, and emitting the DDL without the
check produces exactly the plausible, meaningless result set this PR exists to prevent.

**3. Driver version.** Master was on 1.8.0; this is now 1.9.0, which is both newer and the
first release that has the routing at all. (Worth noting for the record: dropping the patch
without bumping would have resolved to 1.8.0 and silently given a driver with no leader-aware
routing — that could not have happened quietly, since `unstable-strong-consistency` does not
exist in 1.8.0 and cargo fails to resolve.)

### Where the feature is switched on

- `docker build --build-arg CARGO_BUILD_FEATURES=strong-consistency .` — the `Dockerfile`
  takes a build arg that is empty by default, so the ordinary image is byte-for-byte
  unchanged.
- CI runs clippy with the feature both alone and with `user-profile`, a `--profile dist`
  build with it, and the unit tests with it. `strong-consistency-tests` was consuming the
  artifact from `build`, which is now an ordinary binary that refuses `consistency=global`
  outright — that job would have failed on every run for a reason having nothing to do with
  the server — so `build` publishes a second, feature-enabled artifact for it.
- Release artifacts are built **without** the feature.

`--cfg scylla_unstable` is still required by 1.9.0 alongside the crate feature — I checked,
building without it still fails with `field consistency_mode is private` — so the RUSTFLAGS
plumbing stays as it is. It is simply inert now unless the feature is on, and I corrected the
comments in `.cargo/config.toml` and `release.yml` that claimed otherwise: a comment that
overstates why something is there is how it gets deleted by someone tidying up.

### Verified

`cargo fmt --check`; clippy across all six configurations (all targets, `--no-default-features`,
`user-profile`, `strong-consistency`, `strong-consistency` with no default features,
`fetch_extended_version_info`); `--profile dist` builds both with and without the feature; and
the unit tests in both configurations — 85 pass without the feature, 97 with it, plus 19
scylla-bench and 13 library. `test_can_connect` needs a live ScyllaDB and fails without one,
as on master.

Also checked at runtime: the default binary refuses `consistency=global` with the feature
message, and the feature-enabled binary accepts it.

@dawmd — thanks, the release landed right on the estimate.

### Still outstanding

- **Squashing.** The branch is 10 commits. I said I would collapse it once the dependency
  rework landed, and it now has — but I would rather do that as the last step before merge
  than now, so the two commits above stay reviewable against what you last read. Say the word
  and I will squash.
- The `SimpleStrategy` default problem is unrelated to this PR and still needs its own issue.

aleksbykov:
Squashed to a single commit

aleksbykov:
Rebased onto current master (`5edaee0`, which includes #198) and force-pushed — `3c79960` → `3e41149`.

The rebase applied without textual conflicts but needed one real fix: #193 changes `SchemaOption::from_handles` to return `Result<Self>` (it gained a `bail!` for invalid `consistency=` values), while the unit test #198 added was written against the `Self`-returning signature on master. Git merged both happily and the result did not compile; it now `.unwrap()`s.

Also dropped the reviewer note about the `SimpleStrategy` default breaking integration tests — #198 fixed that, so the note was stale. The README caveat about non-tablet keyspaces rejecting `consistency` stays, since that is still true; it just now says `NetworkTopologyStrategy` is the default.

Full matrix re-run green on the rebased head: fmt, clippy in six configurations, `--profile dist` with and without the feature, and the unit tests in both (86 without `strong-consistency`, 98 with — each one up by the test inherited from #198).

aleksbykov:
@CodeLieutenant can you review?

CodeLieutenant:
> @CodeLieutenant can you review?

Yes, of course

CodeLieutenant (approved):

PR #195, opened 2026-09-02
feat: add -errors option to the cassandra-stress frontend

CodeLieutenant:
A longevity test that kills a node with `KillMVBuildingCoordinator` failed the whole benchmark on a `read_failure` error. The replicas of a read are chosen when the read starts, so a replica that dies mid-read makes the coordinator return that error. Every error that survives the retries stops the run, and the retry count was fixed at 9 with no way to change it. A test that kills nodes on purpose reports a false failure.

The `-errors` option accepts `ignore` and `retries=N`, as the original cassandra-stress does. With `ignore`, the worker abandons the operation, records the error in the statistics and continues. The abandoned operation must not keep its cached row, so the row is now bound to the operation id in the write, read, mixed and user workloads.

The integration tests here need #196 to reach the keyspace creation.

Closes QATOOLS-358

### DIFF
--- a/src/bin/cql-stress-cassandra-stress/main.rs
+++ b/src/bin/cql-stress-cassandra-stress/main.rs
@@ -206,6 +206,9 @@ async fn prepare_run(
         }
     };
 
+    let max_retries_per_op = settings.errors.retries;
+    let ignore_errors = settings.errors.ignore;
+
     let operation_factory = create_operation_factory(session, settings, stats).await?;
 
     Ok(Configuration {
@@ -213,8 +216,8 @@ async fn prepare_run(
         concurrency,
         rate_limit_per_second: throttle,
         operation_factory,
-        // TODO: adjust when -errors option is supported
-        max_retries_per_op: 9,
+        max_retries_per_op,
+        ignore_errors,
     })
 }
 
--- a/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
@@ -39,6 +39,7 @@ pub struct MixedOperation {
     clustering_distribution: Box<dyn Distribution>,
     current_operation: MixedSubcommand,
     current_operation_remaining: usize,
+    last_operation_id: Option<u64>,
 }
 
 pub struct MixedOperationFactory {
@@ -81,6 +82,7 @@ impl OperationFactory for MixedOperationFactory {
             clustering_distribution: mixed_params.clustering.create(),
             current_operation: MixedSubcommand::Read,
             current_operation_remaining: 0,
+            last_operation_id: None,
         })
     }
 }
@@ -172,6 +174,12 @@ impl MixedOperation {
             return Ok(ControlFlow::Break(()));
         }
 
+        if self.last_operation_id != Some(ctx.operation_id) {
+            self.last_operation_id = Some(ctx.operation_id);
+            self.cached_row = None;
+            self.current_operation_remaining = self.current_operation_remaining.saturating_sub(1);
+        }
+
         if self.current_operation_remaining == 0 {
             self.current_operation = self.operation_ratio.sample();
             self.current_operation_remaining =
@@ -237,11 +245,6 @@ impl MixedOperation {
                 result
             }
         };
-
-        if result.is_ok() {
-            self.current_operation_remaining -= 1;
-            self.cached_row = None;
-        }
 
         result
     }
--- a/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
+++ b/src/bin/cql-stress-cassandra-stress/operation/mixed.rs
@@ -12,7 +12,6 @@ use cql_stress::{
     make_runnable,
 };
 use scylla::client::session::Session;
-use scylla::value::CqlValue;
 
 use super::{
     counter_write::{CounterWriteOperation, CounterWriteOperationFactory},
@@ -22,7 +21,7 @@ use super::{
     },
     row_generator::RowGenerator,
     write::{WriteOperation, WriteOperationFactory},
-    CassandraStressOperation, CassandraStressOperationFactory, RowGeneratorFactory,
+    CachedRow, CassandraStressOperation, CassandraStressOperationFactory, RowGeneratorFactory,
     DEFAULT_COUNTER_TABLE_NAME, DEFAULT_TABLE_NAME,
 };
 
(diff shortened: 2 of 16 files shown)


soyacz (commented):
## Findings

**1. `-errors` rejects 6 of the 8 suboptions the real tool accepts.** Verified against `SettingsErrors.java` in the Scylla fork: `retries=`, `ignore`, `fail-fast`, `skip-read-validation`, `skip-unsupported-columns`, `delay-policy=`, `min-delay-ms=`, `max-delay-ms=`. This PR implements two and makes the other six a hard parse error — `cs_args_bad_test.in:63` codifies that for `skip-unsupported-columns`. This repo already has a convention for exactly this case: `NotSupported` (`param/types.rs:141`), used across `transport.rs` to accept-and-warn, so `help -transport` lists the suboption and marks it "(unsupported)". Apply the same to the six. A command copied from a real c-s invocation currently aborts at startup.

  Mitigating, and worth knowing: SCT probes with `cassandra-stress help -errors | grep "^Usage:"` and only injects `skip-unsupported-columns` if the Usage line lists it (`sdcm/stress_thread.py:208`). `Usage: -errors [ignore] [retries=?]` parses correctly there, so the automatic SCT path is safe. Hand-written test-case yamls use only `ignore` and `retries=`, which work.

**2. The riskiest part of the diff has no test.** `mixed.rs` did more than move the row reset: `current_operation_remaining` moved from a post-success `-= 1` to a pre-execution `saturating_sub(1)`. That changes which subcommand runs and how long a clustering burst lasts. I traced it and it is correct — the burst still runs exactly N operations — but the only new unit test covers `GenericCassandraStressOperation`, the simplest of the three copies. `MixedOperation` and `UserOperation` got the same new state machine with zero coverage. Add a mixed-workload test that drives two ids through a burst boundary.

  Related: `saturating_sub` is required for the first call (remaining starts at 0), but it also removes the debug-build underflow panic that was the only signal the counter invariant broke.

**3. `last_operation_id` bookkeeping is copy-pasted three times** — `operation/mod.rs:104`, `mixed.rs:177`, `user.rs:104`. Nothing forces a future `Operation` that caches rows to repeat it, and getting it wrong is silent. A small `CachedRow { row, last_operation_id }` with a `get_or_generate(ctx, f)` method would carry the invariant in one place.

**4. The compose image change is unrelated and its justification is wrong.** The commit says it "pins the LTS image the tests run against". `scylladb/scylla:2026.1` is a stream tag, not a pin — it was last pushed 2026-09-01. Tablets are enabled by default in 2026.1 as well, so the pin is not what fixed `SimpleStrategy`; the `NetworkTopologyStrategy` change is. The effect is that CI stops testing 2026.2 and 2026.3. Either drop it, or keep it and say in the commit body what actually broke on `latest`.

## Nits

- `tools/test_cs_errors_ignore.py:83` — `if not total_errors` treats "line missing" and "zero errors" the same, and the message prints `Got: None`. Split the two. The failing-run branch prints `stdout` but not `stderr`.
- `errors.rs:7` — `DEFAULT_RETRIES_PER_OPERATION: u64` cast `as usize` twice. Parse straight into the field type.
- Java c-s prints `Tries: 10` (retries + 1); this prints `Retries: 9`. The retry arithmetic itself matches exactly (`tries = max(1, retries + 1)` = 10 attempts, same as `trial_idx >= 9`).


Also worth stating in the PR: with `ignore`, an entirely unreachable cluster now exits 0. That is c-s parity, and there is no error-limit escape hatch on either side.

docker/scylla-test/compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   scylla-test:
-    image: scylladb/scylla
+    image: scylladb/scylla:2026.1
```
fruch: pining it, hints there an issue with running newer version, are track it ? attend to it ?
(also it's not part of any of the commits mentioned in the PR)

fruch: this should be parameterize, and controlled from the github action, so we can somehow move back and forth in PR, temporarily if urgent/needed

tools/util/cassandra_stress.py
```diff
@@ -35,7 +35,8 @@ def prepare_args(command, node_ip, keyspace, runtime_args: CSCliRuntimeArguments
         rate_args.append(f"throttle={runtime_args.throttle}")
     args.extend(["-rate"] + rate_args)
     
-    args.extend(["-schema", f"keyspace={keyspace}"])
+    args.extend(["-schema", "replication(strategy=NetworkTopologyStrategy)",
```
fruch, line 38: isn't that something getting fixed in c-s ? maybe worth a note with ref to issue or PR, so it would be remove later

fruch (approved):
LGTM

even that I really hate how golang (or our repos) mixes tests with implementation, it really confusing to me to review

docker/scylla-test/compose.yml
CodeLieutenant: I'm pinning it as a test, there is no indication of an error, all of our tools have scylla version pinned in tests, so I did the same, also it surfaced the problem with SimpleReplication, which is not supported by scylla anymore.

tools/util/cassandra_stress.py
CodeLieutenant: No, it's not getting fixed in c-s, at least I dont know about that. `replication(strategy=NetworkTopologyStrategy)` here is a must, because scylla does not support the Simple anymore, and the default is Simple still.

fruch: those test are no using our c-s, right ?

docker/scylla-test/compose.yml
fruch: o.k. where is the followup jira/PR, that would move this pin into lastest scylla ? and address the problems ?

CodeLieutenant: We dont have problems on running latest scylla. We just never did it like do, should we?

I can revert this change

tools/util/cassandra_stress.py
CodeLieutenant: They are, it compares the results from c-s with the cql-stress

docker/scylla-test/compose.yml
fruch: pinning it, with no process/code in place that would update, it would be like that for longer then needed, and might miss important changes in scylla, that need attention 

if it's keep like this, we get breakage unaccepted in unrelated PRs

if one of those two, I prefer the 2nd.

a more controlled way is to have it pinned, but make sure renovate would send us PR for updates about it (I hope it's identifying docker-compose files), and then we have best of both worlds (if it's working)

tools/util/cassandra_stress.py
fruch: so now I understand this is the actual fix for the issue with 2026.2+

I still argue it should go separately then this PR, as it was noticed in multiple other PRs

docker/scylla-test/compose.yml
CodeLieutenant: We can use renovate to update scylla versions as well, pin them to the latest right now, 2026.2 since it's the latest stable, make renovate send PR's for upgrading?

tools/util/cassandra_stress.py
CodeLieutenant: I argue against it, as if I dont see passing tests I will not merge it, and it will delay this PR, and it's already working.

docker/scylla-test/compose.yml
fruch: Yes we can, and still all of that changes and confirmation that renovate would work as expected, should happen in its own PR

tools/util/cassandra_stress.py
fruch: I think you have enough memory/context to remember that it's worked, take a commit out and then push.

If we know exactly what is wrong in a PR, and that it going to be dealt with, there no problem merging a PR.

